PR #3, opened 2016-02-16
L5pc test fix

mgeplf:
* can be run from any path

### DIFF
--- a/bluepyopt/tests/test_l5pc.py
+++ b/bluepyopt/tests/test_l5pc.py
@@ -4,11 +4,11 @@ import os
 
 import nose.tools as nt
 
-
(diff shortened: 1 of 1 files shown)


bluepyopt/tests/test_l5pc.py
```diff
@@ -4,11 +4,11 @@
 
 import nose.tools as nt
 
-
-l5pc_dir = '../../examples/l5pc'
+L5PC_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__),
```
wvangeit: At the moment this won't work anyway, because the morphology path and mod library still need to be in this directory.

PR #4, opened 2016-02-16
put the destroy() method in the base classes

mgeplf:
* so don't need to repeat one's self

### DIFF
--- a/bluepyopt/electrical/objectives.py
+++ b/bluepyopt/electrical/objectives.py
@@ -49,6 +49,7 @@ class SingletonObjective(Objective):
 
         Objective.__init__(self, name, [feature])
 
+    @property
     def calculate_score(self, responses):
         """Objective score"""
 
--- a/bluepyopt/electrical/parameters.py
+++ b/bluepyopt/electrical/parameters.py
@@ -93,6 +93,10 @@ class Parameter(object):
                     (self.name, self._value, str(self.lower_bound),
                      str(self.upper_bound)))
 
+    def destroy(self):
+        """Destroy parameter instantation"""
+        pass
+
 
 class NrnGlobalParameter(Parameter):
 
@@ -129,10 +133,6 @@ class NrnGlobalParameter(Parameter):
             return '%s: %s = %s' % (self.name, self.param_name, self.value)
         else:
             return '%s: %s = %s' % (self.name, self.param_name, self.bounds)
-
-    def destroy(self):
-        """Destroy parameter instantation"""
-        pass
 
 
 class NrnSectionParameter(Parameter):
@@ -179,10 +179,6 @@ class NrnSectionParameter(Parameter):
                 self.param_name,
                 location,
                 self.value)
-
-    def destroy(self):
-        """Destroy parameter instantation"""
-        pass
 
     def __str__(self):
         """String representation"""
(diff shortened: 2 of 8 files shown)


bluepyopt/electrical/objectives.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, name, feature):
 
         Objective.__init__(self, name, [feature])
 
+    @property
```
wvangeit: ?

mgeplf: Yeah, was just going to comment on that.

MaxObjective and WeightedSumObjective both wrap that method w/ '@property', so to keep the same interface, I think it should have it too

bluepyopt/electrical/parameters.py
```diff
@@ -93,6 +93,10 @@ def check_bounds(self):
                     (self.name, self._value, str(self.lower_bound),
                      str(self.upper_bound)))
 
+    def destroy(self):
```
wvangeit: Nope, this shouldn't be here.
'Parameters' still has to move a level up. 'destroy() / initiate()' will be something specific to Neuron (and maybe other simulators). But you can already replace this Parameter with NrnParameter in the mean time if you want.

bluepyopt/tests/test_bluepyopt.py
```diff
@@ -33,4 +33,4 @@ def test_import():
 def test_neuron_import():
     """BluePyOpt: test if bluepyopt.neuron import was successful"""
     import bluepyopt.importer  # NOQA
-    nt.assert_is_instance(bluepyopt.importer.neuron, types.ModuleType)
+    nt.ok_(isinstance(bluepyopt.importer.neuron, types.ModuleType))
```
wvangeit: what's the advantage ?

bluepyopt/electrical/parameters.py
mgeplf: Wouldn't a consistent interface make it easier so one doesn't always have to check if a method exists? A no-op destroy reduces the differences between Parameter subclasses

bluepyopt/tests/test_bluepyopt.py
mgeplf: assert_is_instance doesn't exist in all nose versions, since it's pulled in from unittest.TestCase

bluepyopt/electrical/parameters.py
wvangeit: Don't agree. If you pass an object to another object that will call 'destroy()' on it, but that object doesn't naturally support it, you're very probably doing something wrong. Python should then throw an exception instead of being silent.

bluepyopt/tests/test_bluepyopt.py
wvangeit: you mean that it doesn't work in 2.6 ? 
Would rather call that a feature instead of a bug ;-)
But anyway, we can change it

bluepyopt/electrical/parameters.py
mgeplf: Ok, does that apply only to parameters?

For example, bluepyopt/electrical/protocols.py calls destroy on all stimulus & recordings.  I moved their destroys up to the base class, so knows they have to implement it, otherwise there's no hint as to which methods are needed to be a valid stimulus.

The alternative, imo, is to use ABCs such that if a method isn't defined, but is needed, then the class can't be instantiated.

wvangeit: I think we should wait with all these details until the API is stable.
Like I said I still have to move around Parameters.py (up a level), and other more important API changes are in the pipeline. (All these subclasses of Parameters are probably going to disappear and replaced by NrnParameter, and a new class called Variables is going to be introduced).

mgeplf: Ok, I'll close this up then

PR #10, opened 2016-02-18
Added seed argument to Optimisation

wvangeit:
Disabling 1 instance check in Optimisation for the moment because it
clashes with scoop


bluepyopt/optimisation.py
```diff
@@ -123,7 +129,7 @@ def __lt__(self, other):
             fitness=deap.creator.WeightedSumFitness)
 
         # Set random seed
-        random.seed(1)
+        random.seed(self.seed)
```
mgeplf: looks good to me

PR #14, opened 2016-02-18
simplecell test using scoop

mgeplf:


mgeplf:
Let's see if this passes CI, or if I the output for the best individual is different

mgeplf:
huh, neat

wvangeit:
Ok, thanks

PR #17, opened 2016-02-23
Destroy tvector in recording, didn't happen before

wvangeit:


wvangeit:
Strange, this bug generate pickle error on CSCS viz. Strange that this happen on any other machine

PR #52, opened 2016-03-09
add unzip and matplotlib to installs

mgeplf:
* Had to add some dependencies to the debian/ubuntu package
  list so that matplotlib will build


wvangeit:
you don't want this merged anymore ?

mgeplf:
I changed the hosts by accident when I was testing, plus I want to work on a branch

PR #53, opened 2016-03-09
add unzip and matplotlib to installs

mgeplf:
* Had to add some dependencies to the debian/ubuntu
  package list so that matplotlib will build


wvangeit:
Might want to sync this with master

mgeplf:
it's sync'd

PR #62, opened 2016-04-13
have l5pc example create x86_64 link for compiled mechanisms

mgeplf:


examples/l5pc/opt_l5pc.py
```diff
@@ -76,10 +76,11 @@ def main():
         logger.debug('Doing compile')
         import commands
         commands.getstatusoutput('cd mechanisms/; nrnivmodl; cd ..')
-
-    # TODO store definition dicts in json
```
mgeplf: I removed these, b/c they are the same as at the top of the file

examples/l5pc/opt_l5pc.py
```diff
@@ -76,10 +76,11 @@ def main():
...
-    # TODO rename 'score' into 'objective'
-    # TODO add functionality to read settings of every object from config format
+        #NEURON expects the compiled mechanisms to be in ./x86_64
```
wvangeit: Actually, I don't think we need the 'cd' command.
We just need 'nrnivmodl mechanisms' in the root

PR #66, opened 2016-04-20
Include IBEA code inside bluepyopt

wvangeit:
Compiler now became dependency
New subpackage 'deapext'
Backward compatibility is maintained for the moment


bluepyopt/deapext/tools/eps.c
```diff
@@ -0,0 +1,86 @@
...
+ The authors were Werner Van Geit, Michael Gevaert and Jean-Denis Courcol       
+ It is based on a C implementation of the IBEA algorithm in the PISA            
+ optimization framework developed at the ETH, Zurich                            
+ http://www.tik.ee.ethz.ch/pisa/selectors/ibea/?page=ibea.php                   
+ 
+*/
+
+#include <Python.h>
+
+static PyObject * eps_indicator(PyObject *self, PyObject *args)
```
mgeplf: do you have benchmarks on the speedup from this?

wvangeit: Would have to look it up, was something I did a year ago.
As far as I know I sped up Giuseppe's code in 3 stages, for in total a speedup of factor 800 or so.
10x was getting rid of numpy, 10x was preventing some copies in memory and the last part was this C code. 
We should anyway add some tests to this code, and we might want to compare to a python implementation.
We could also run Giuseppe synaptic plasticity code with/without it.

PR #71, opened 2016-04-26
Mgeplf sel ibea

wvangeit:


bluepyopt/deapext/tools/selIBEA.py
```diff
@@ -68,52 +54,37 @@ def selIBEA(population, mu, alpha=None, kappa=.05, tournament_n=4):
...
+    # Calculate minimal square bounding box of the objectives
+    box_ranges = (numpy.max(population_matrix, axis=0) -
+                  numpy.min(population_matrix, axis=0))
+
+    components_matrix = numpy.zeros((pop_len, pop_len))
+    for i in xrange(0, pop_len):
+        diff = population_matrix - population_matrix[i, :]
+        components_matrix[i, :] = numpy.max(
+            numpy.divide(
+                diff,
```
wvangeit: @mgeplf, you removed the indicator function here. We need to be able to switch this out with e.g. the hypervolume indicator. Is there a way to keep this indicator function call ?

wvangeit: @mgeplf, here is the hypervolume indicator https://github.com/DEAP/deap/blob/master/deap/tools/indicator.py. We still have to decide which one has the highest performance.

mgeplf: I didn't realize the indicator was a parameter.  I will have a look at that.

mgeplf: How would the code have been modified to call that indicator?  It's probably best to have a 
`_calc_fitness_components_eps` and `_calc_fitness_components_hypervolume`? and then `selIBEA()` can call whichever the user wants?

bluepyopt/deapext/tools/selIBEA.py
```diff
@@ -68,52 +54,37 @@ def selIBEA(population, mu, alpha=None, kappa=.05, tournament_n=4):
...
-    max_absolute_indicator = numpy.max(abs(components_matrix))
+    max_absolute_indicator = numpy.max(numpy.abs(components_matrix))
 
     # Normalisation
-    components_matrix = \
-        numpy.exp(numpy.multiply(components_matrix,
-                                 -1.0 / (kappa * max_absolute_indicator)))
-
+    components_matrix = numpy.exp((-1.0 / (kappa * max_absolute_indicator)) *
+                                  components_matrix.T)
```
wvangeit: why do we need the transpose here ?

bluepyopt/deapext/tools/selIBEA.py
```diff
@@ -68,52 +54,37 @@ def selIBEA(population, mu, alpha=None, kappa=.05, tournament_n=4):
...
-
-    # pylint: disable=F0401, E0611
-    import eps
-    # pylint: enable=F0401, E0611
-
-    # Calculator the indicator value for every element in the matrix
-    # The code inside this for loop is (has to be) heavily optimised for speed
-    for i in xrange(0, pop_size):
-        ind1 = population_matrix[i]
-        for j in itertools.chain(xrange(0, i), xrange(i + 1, pop_size)):
```
wvangeit: @mgeplf how did you implement the skipping of the i-th element in your matrix operation ?

bluepyopt/deapext/tools/selIBEA.py
mgeplf: To match the original output format.  Alternatively, I can transpose every operation, but it was faster to just do it in bulk at once.

bluepyopt/deapext/tools/selIBEA.py
mgeplf: The diagonals being zero are handled by the fact that the subtractions are operating on themselves.  Before, I think you could have done the same thing: b/c when i == j, then ind1 - ind2 == 0

bluepyopt/deapext/tools/selIBEA.py
mgeplf: In other words, in the old code:
```
    for i in xrange(0, pop_size):
        ind1 = population_matrix[i]
        for j in itertools.chain(xrange(0, i), xrange(i + 1, pop_size)):
            ind2 = population_matrix[j]
            components_matrix[i, j] = eps.indicator(ind1,
                                                    ind2,
                                                    min_box_bounds,
                                                    max_box_bounds)
```
is equivalent to:
```
    for i in xrange(0, pop_size):
        ind1 = population_matrix[i]
        for j in xrange(0, pop_size):
            ind2 = population_matrix[j]
            components_matrix[i, j] = eps.indicator(ind1,
                                                    ind2,
                                                    min_box_bounds,
                                                    max_box_bounds)
```
I'm just exploiting that fact in a vectorized way.

bluepyopt/deapext/tools/selIBEA.py
mgeplf: In other words, in the old code:
```
    for i in xrange(0, pop_size):
        ind1 = population_matrix[i]
        for j in itertools.chain(xrange(0, i), xrange(i + 1, pop_size)):
            ind2 = population_matrix[j]
            components_matrix[i, j] = eps.indicator(ind1,
                                                    ind2,
                                                    min_box_bounds,
                                                    max_box_bounds)
```
is equivalent to:
```
    for i in xrange(0, pop_size):
        ind1 = population_matrix[i]
        for j in xrange(0, pop_size):
            ind2 = population_matrix[j]
            components_matrix[i, j] = eps.indicator(ind1,
                                                    ind2,
                                                    min_box_bounds,
                                                    max_box_bounds)
```
I'm just exploiting that fact in a vectorized way.

PR #73, opened 2016-04-27
Allow debug logging to be turned off

mgeplf:
* a 'run_fit.py --start' takes about 25% longer with logging.debug
  turned on.  Thus, add a wrapper to the logging.debug functions
  so that it can be turned off.
* added a wrapper for printing vectors, so that expanded string is
  only created if it is needed

### DIFF
--- a/examples/graupnerbrunelstdp/stdputil.py
+++ b/examples/graupnerbrunelstdp/stdputil.py
@@ -15,6 +15,22 @@ from scipy.special import erf
 
 logging.basicConfig(level=logging.WARN)
 
+#Note: having debug logging statements increases the run time by ~ 25%,
+# because they exist in tight loops, and expand their outputs, even when
+# debug is off, so we disable logging if possible.  Set this to true if
+# verbose output is needed
+LOGGING_DEBUG = False
+
+def logging_debug_vec(fmt, vec):
+    '''log to debug a vector'''
+    if LOGGING_DEBUG:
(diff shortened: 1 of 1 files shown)


examples/graupnerbrunelstdp/stdputil.py
```diff
@@ -15,6 +15,22 @@
 
 logging.basicConfig(level=logging.WARN)
 
+#Note: having debug logging statements increases the run time by ~ 25%,
+# because they exist in tight loops, and expand their outputs, even when
+# debug is off, so we disable logging if possible.  Set this to true if
```
wvangeit: so if I understand this correctly, we should get rid of the logging module everywhere ?
Probably should implement a printv() function or something.

mgeplf: No, that's not the take away at all.  It's more that we have to be careful about functions that are called many, many times (like logging.debug() in `time_above_threshold`)

wvangeit: but you're saying that they logging.debug executes code even when the debugging is turned off ?

mgeplf: Yeah, depending on what the logging level is, it does execute some code.  It bails out relatively early, but there is some overhead.

It looks like half the speedup comes from the logging_debug_vec not expanding the arrays, and the other half from forcing debug to be turned off in the tight loop.

PR #74, opened 2016-04-27
Fix typo in NrnSeclistSecLocation

mgeplf:
* move common sectionlist code to function
* Use super() style inheritence calls
* Cleaned up some typo

### DIFF
--- a/bluepyopt/ephys/locations.py
+++ b/bluepyopt/ephys/locations.py
@@ -19,12 +19,13 @@ Copyright (c) 2016, EPFL/Blue Brain Project
  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """
 
+import itertools
 from .importer import neuron
 
 
 class Location(object):
 
-    """EPhys feature"""
+    """Location"""
 
     def __init__(self, name):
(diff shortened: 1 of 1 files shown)


bluepyopt/ephys/locations.py
```diff
@@ -37,6 +38,19 @@ def __init__(self, name):
 # TODO specify in document abrevation comp=compartment, sec=section, ...
 
 
+def _nth_isectionlist(isectionlist, index):
```
mgeplf: what's the expected behavior if len(sectionlist) == 10, and index >= 10?

bluepyopt/ephys/locations.py
```diff
@@ -195,5 +188,5 @@ def instantiate(self, sim=None, icell=None):
     def __str__(self):
         """String representation"""
 
-        return '%f mum from soma in %s' % (
+        return '%f um from soma in %s' % (
```
mgeplf: just to make sure: the unit is micrometer here, right?

wvangeit: yeah, but mum is ok no. Otherwise micron would be better.

PR #76, opened 2016-04-28
Add master dt and cvode_active to simulator class

croessert:
This change makes it possible to declare a master dt and cvode_active in the simulator class

### DIFF
--- a/bluepyopt/ephys/simulators.py
+++ b/bluepyopt/ephys/simulators.py
@@ -1,9 +1,5 @@
 """Simulator classes"""
 
-import os
-import imp
-import ctypes
-
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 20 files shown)


bluepyopt/ephys/simulators.py
```diff
@@ -22,17 +22,34 @@ def __init__(self):
...
 
         self.neuron.h.tstop = tstop
+
+        if cvode_active is None: # use cvode_active of simulator
+            cvode_active = self.cvode_active
+
         self.neuron.h.cvode_active(1 if cvode_active else 0)
-        logger.debug('Running Neuron simulator %.6g ms', tstop)
+
+        if dt is None: # use dt of simulator
```
wvangeit: I think we might want to raise an exception if the user specifies cvode_active AND dt. What do you think ?
Shouldn't we also get the default dt from Neuron, instead of hard coding it to 0.025 ?

wvangeit:
Am closing this pull request. Is now replaced by https://github.com/BlueBrain/BluePyOpt/pull/85
(This branch became a bit of a mess after I did a rebase)

PR #80, opened 2016-04-28
make sure all ephys classes have __str__

mgeplf:


bluepyopt/ephys/locations.py
```diff
@@ -133,7 +133,7 @@ def instantiate(self, sim=None, icell=None):
     def __str__(self):
         """String representation"""
 
-        return '%s[%s]' % (self.seclist_name, self.sec_index)
+        return '%s: %s[%s]' % (self.name, self.seclist_name, self.sec_index)
```
wvangeit: As far as I remember, I designed these strings, so that they could be used by higher classes to form nice sentences. I don't think that will work after this change anymore ?

mgeplf: It's mainly adding the name that's stored by the base class, so it shouldn't change the structure that much.  Is there a particular example that will change the view?

wvangeit: e.g https://github.com/BlueBrain/BluePyOpt/blob/master/bluepyopt/ephys/parameters.py#L183

mgeplf: Maybe it should be the old format string, but with (self.name, self.sec_index).  I think the confusion comes from there being both a self.name and a self.sec_list.  Right now, if someone has multiple locations that use sec_list = 'somatic', they don't know which one is being set.

wvangeit: well, the only information contained in this location is the sec_list and sec_index. It doesn't matter if it's a different object setting, the user has all the information necessary

mgeplf: Location also has a name, though.  Should name be set to sec_name?

wvangeit: No, because user should be able to freely choose name

mgeplf: Ok, but in that case, shouldn't that name get displayed when the user calls str() on the object?

wvangeit: nope

mgeplf: But that seems like a relevant piece of information.  One could also define two locations with different names, but the same values, and then tracking down which one emitted that string won't be possible.

PR #87, opened 2016-05-02
ephys Parameter now inherits from main Parameter

wvangeit:


bluepyopt/ephys/parameters.py
```diff
@@ -19,95 +19,47 @@
...
+            value=value,
+            frozen=frozen,
+            bounds=bounds)
+
+    @abstractmethod
+    def instantiate(self, sim=None, icell=None):
+        """Instantiate the parameter in the simulator"""
         pass
 
-    def raise_if_no_value(self):
```
trialsbiker: why not keep this so the code doesn't have to be duplicated in each function?

mgeplf: why not keep this so the code doesn't have to be duplicated each time?  If you want the class name, just grab self.__class__.__name__

wvangeit: It's only two lines. Find it much more readable to explicitly code it than adding another function call. Other parameter types could also have different requirements.

mgeplf: But it also means one has to write more tests across this error path, and when there's an update, it has to be applied in multiple places.

PR #88, opened 2016-05-02
Update codecov.yml

stevepeak:
Needed the context for the status. Thanks!


wvangeit:
@stevepeak thanks for the fix !

PR #89, opened 2016-05-03
say which paramemeter isn't frozen when instantiating

mgeplf:


bluepyopt/ephys/parameters.py
```diff
@@ -128,8 +128,8 @@ def instantiate(self, sim=None, icell=None):
         """Instantiate"""
         if self.value is None:
             raise Exception(
-                'NrnSectionParameter: impossible to instantiate parameter '
-                'without value')
+                'NrnSectionParameter: impossible to instantiate parameter "%s" '
```
mgeplf: re-collapsing this logic back into a helper method is probably still a good idea

PR #97, opened 2016-05-08
Fixed broken analysis in the graupnerbrunel example

pineider:


pineider:
The analysis was failing because of a missing pickle file (cache). I added a check on the existence of the file, but maybe we should just recompute every time. I didn't notice any speed-up when using the cache, probably because most of the time is spent generating the figures rather than computing the data. Furthermore, the cache should be invalidated if the user changes any optimization  parameters, but we cannot detect these situations.

PR #98, opened 2016-05-09
remove bbp version of DEAP from requirements section

jdcourcol:


wvangeit:
In the end, I think we should still put DEAP in the list. But the title of that section should probably be 'dependencies'

jdcourcol:
and ipyparallel

PR #106, opened 2016-06-02
Add serializer 'DictMixin' for ephys classses

mgeplf:
* can .to_dict() a most ephys classes, returns a dictionary
  that can be turned into json
* recursively will serialize members that have the to_dict()
  member
* can deserialize arbitrary dicts (say, from json), that were
  serialized using the 'instantiator'

### DIFF
--- a/bluepyopt/ephys/efeatures.py
+++ b/bluepyopt/ephys/efeatures.py
@@ -22,6 +22,7 @@ Copyright (c) 2016, EPFL/Blue Brain Project
 
 import logging
 
+from bluepyopt.ephys.serializer import DictMixin
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


bluepyopt/ephys/efeatures.py
```diff
@@ -22,6 +22,7 @@
 
 import logging
 
+from bluepyopt.ephys.serializer import DictMixin
```
wvangeit: I'm wondering, should this be ephys-specific ?

mgeplf: Up to you.  Are the other classes going to be serialized?  Won't they be more complicated, if they are?

I still think it might be worthwhile to break out the ephys portion, so that's another reason to leave it here.

I'm agnostic on the issue.

wvangeit: Mmm, I guess we can leave it here for now, probably fairly easy to change afterwards. I would be fairly 'cool' if we can read/write and entire optimisation configuration this way

mgeplf: That's something we'd have to test.

I still want to write tests for all the other classes that serialize, to make sure I didn't miss anything.

I can do that in another PR though, if you're happy w/ this one.

mgeplf: We'll probably need some more helper functions for serializing groups of objects, but it's basically just a list comprension: `[o.to_dict() for o in objects]`

PR #108, opened 2016-06-03
mock package is only needed for testing

mgeplf:


setup.py
```diff
@@ -28,7 +28,7 @@
     version=versioneer.get_version(),
     cmdclass=versioneer.get_cmdclass(),
     install_requires=['numpy>=1.6', 'pandas>=0.18', 'deap', 'efel>=2.6',
-                      'scoop>=0.7', 'ipyparallel', 'mock'],
+                      'scoop>=0.7', 'ipyparallel', ],
```
wvangeit: does mock take a long time to install ? Would just leave that in.

mgeplf:
I think it's best practice to keep the dependencies low.  Since it's only needed for testing, it's handled the same way that nose & coverage are.

PR #109, opened 2016-06-03
Add comment field for serializable models

mgeplf:
* allow's one to document why certain choices were made
* moved name & comment into a base class


mgeplf:
I didn't do 'parameters', b/c its inheritance pattern is different, and name is stored in NrnParameter(bluepyopt.parameters.Parameter.  This may be worth looking into.

PR #123, opened 2016-06-24
Added support for elitism

wvangeit:


wvangeit:
@pineider should we push this in the master ?

PR #129, opened 2016-07-05
minor fixes for Windows

Chris-crisur:
hoc.so does not exist on NEURON Windows, so a check is done for the
platform system before processing the file.
directing hoc stdout to '/dev/null' fails as it doesn't exist on Windows, so it was replaced with 'null'

### DIFF
--- a/bluepyopt/ephys/morphologies.py
+++ b/bluepyopt/ephys/morphologies.py
@@ -94,7 +94,7 @@ class NrnFileMorphology(Morphology, DictMixin):
         # TODO this is to get rid of stdout print of neuron
         # probably should be more intelligent here, and filter out the
         # lines we don't want
-        sim.neuron.h.hoc_stdout('/dev/null')
+        sim.neuron.h.hoc_stdout('null')
         imorphology.input(str(self.morphology_path))
         sim.neuron.h.hoc_stdout()
 
--- a/bluepyopt/ephys/morphologies.py
+++ b/bluepyopt/ephys/morphologies.py
@@ -22,6 +22,7 @@ Copyright (c) 2016, EPFL/Blue Brain Project
 # pylint: disable=W0511
 
 import os
+import platform
 import logging
 from bluepyopt.ephys.base import BaseEPhys
 from bluepyopt.ephys.serializer import DictMixin
@@ -94,7 +95,12 @@ class NrnFileMorphology(Morphology, DictMixin):
         # TODO this is to get rid of stdout print of neuron
         # probably should be more intelligent here, and filter out the
         # lines we don't want
-        sim.neuron.h.hoc_stdout('NUL')
+
+        if platform.system()=='Windows':
+            sim.neuron.h.hoc_stdout('NUL') 
+        else:
+            sim.neuron.h.hoc_stdout('/dev/null') 
+
         imorphology.input(str(self.morphology_path))
         sim.neuron.h.hoc_stdout()
 
--- a/bluepyopt/ephys/morphologies.py
+++ b/bluepyopt/ephys/morphologies.py
@@ -94,7 +94,7 @@ class NrnFileMorphology(Morphology, DictMixin):
         # TODO this is to get rid of stdout print of neuron
         # probably should be more intelligent here, and filter out the
         # lines we don't want
-        sim.neuron.h.hoc_stdout('null')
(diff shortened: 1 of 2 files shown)


wvangeit:
@Chris-crisur, thanks for contributing these changes. Sorry for the late reply, was at a conference. Will have a look at it as soon as possible

bluepyopt/ephys/morphologies.py
```diff
@@ -94,7 +94,7 @@ def instantiate(self, sim=None, icell=None):
         # TODO this is to get rid of stdout print of neuron
         # probably should be more intelligent here, and filter out the
         # lines we don't want
-        sim.neuron.h.hoc_stdout('/dev/null')
+        sim.neuron.h.hoc_stdout('null')
```
wvangeit: If I look at the Neuron documentation, it seems this will create a file called 'null' ?
Maybe we should put a switch statement here, to make sure that /dev/null is still used on all systems apart from Windows.

mgeplf: Maybe NUL would work? (I don't have a windows machine)
It's a special file in windows that's somewhat equivalent to /dev/null, iirc.
https://technet.microsoft.com/library/Cc961816

wvangeit: Yeah, was thinking about that too. @Chris-crisur, would you mind trying that out (we don't have access to Windows machines here) ?

bluepyopt/ephys/simulators.py
```diff
@@ -16,11 +16,14 @@ def __init__(self, dt=None, cvode_active=True, cvode_minstep=None):
...
         import ctypes
-
-        hoc_so = os.path.join(imp.find_module('neuron')[1] + '/hoc.so')
-
-        nrndll = ctypes.cdll[hoc_so]
-        ctypes.c_int.in_dll(nrndll, 'nrn_nobanner_').value = 1
+        import platform
+
+        if platform.system()!='Windows':
+            # hoc.so does not exist on NEURON Windows
```
wvangeit: Hi @nrnhines, is there a construct to disable the Neuron banner, similar to the code below, that works on Windows ?

nrnhines: > is there a construct to disable the Neuron banner, similar to the code below, that works on Windows ?
 On launch you can use the option -nobanner
(see all options with:     nrniv -h )

wvangeit: Mmm, we would need a construct that works after starting python. Typically the bluepyopt scripts are not run from an 'nrniv -python' pompt.

bluepyopt/ephys/morphologies.py
Chris-crisur: tested:
"null" created a file (as mentioned
"NUL" does not create a file, no errors given; seems to work

wvangeit: @Chris-crisur Ok, great. Do you mind changing that in your branch.
I could also do it, but then I have to create a new PR.

bluepyopt/ephys/morphologies.py
```diff
@@ -94,7 +94,7 @@ def instantiate(self, sim=None, icell=None):
         # TODO this is to get rid of stdout print of neuron
         # probably should be more intelligent here, and filter out the
         # lines we don't want
-        sim.neuron.h.hoc_stdout('/dev/null')
+        sim.neuron.h.hoc_stdout('NUL')
```
wvangeit: Sorry, we actually need to make this system dependent, otherwise it won't work on Linux

bluepyopt/ephys/morphologies.py
```diff
@@ -94,7 +95,12 @@ def instantiate(self, sim=None, icell=None):
         # TODO this is to get rid of stdout print of neuron
         # probably should be more intelligent here, and filter out the
         # lines we don't want
-        sim.neuron.h.hoc_stdout('NUL')
+
+        if platform.system()=='Windows':
```
Chris-crisur: @wvangeit sorry about skimping on platform-dependence.
hopefully this clears it up. my bad.

also, may need to write more unit tests for this class too as travis didn't fail for the previous commit :/

wvangeit: k, thanks for the fix. 
Yeah, it ran on Linux, but it must have created a 'NUL' file

PR #132, opened 2016-07-14
make sure latest version of bluepyopt is installed

mgeplf:


cloud-config/roles/deap/tasks/main.yaml
```diff
@@ -22,14 +22,15 @@
 - name: Install scoop
   pip: name=scoop version="{{ scoop_version }}" virtualenv={{ venv }}
 
-- name: Install scoop
```
mgeplf: apparently I was installing scoop twice? weird

mgeplf:
going to commit this so Elisabetta can try it

PR #139, opened 2016-07-29
Create NEURON hoc files from ephys models

mgeplf:
* initial version, considered beta
* given a set of mechanisms, and frozen parameters, will
  generate a hoc cell template that can be loaded in neuron,
  and instantiated with a morphology.
* Much of the template hoc code comes from
    https://bbp.epfl.ch/nmc-portal/downloads
  But was refactred to be self contained and simpler


mgeplf:
I also want to get a test in that checks that a results of loading a hoc file and running a protocol on it is the same as setting up a cell using BluePyOpt, so that we can be sure it's creating exactly the same code.  Hopefully I will get to that soon.

wvangeit:
So I can merge ?

wvangeit:
Actually, one question, why is this the create_hoc function not part of the CellModel class ?

wvangeit:
And how do we specify the morphology ?

wvangeit:
We actually need an API like: model.create_hoc(param_values, template_name=, template=)
I guess it just involves adding a method to model that calls the existing ones.
(I would actually prefer if all this functionality is made part of the CellModel class)

mgeplf:
* I can put it in CellModel
* It's a cell template, so morphology would be specified on instantiation
* I can make it behave like that, no problem

wvangeit:
Ok. Understand about morphology. It would still be nice to have template that knows about the morphology name. Maybe we can make it an option ?

mgeplf:
I actually have that, and purposely removed it :)
I'll add it back in.

PR #160, opened 2016-09-26
Stoch general

croessert:
Making setting of stochastic rng more general to allow for other channels with keyword "Stoch".

### DIFF
--- a/bluepyopt/ephys/mechanisms.py
+++ b/bluepyopt/ephys/mechanisms.py
@@ -107,7 +107,7 @@ class NrnMODMechanism(Mechanism, DictMixin):
     def instantiate_determinism(self, deterministic, icell, isec, sim):
         """Instantiate enable/disable determinism"""
 
-        if self.prefix == 'StochKv':
+        if 'Stoch' in self.prefix:
             setattr(
                 isec,
                 'deterministic_%s' %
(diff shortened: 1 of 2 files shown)


bluepyopt/ephys/mechanisms.py
```diff
@@ -119,16 +119,16 @@ def instantiate_determinism(self, deterministic, icell, isec, sim):
                 short_secname = sim.neuron.h.secname(sec=isec).split('.')[-1]
                 for iseg in isec:
                     seg_name = '%s.%.19g' % (short_secname, iseg.x)
-                    sim.neuron.h.setdata_StochKv(iseg.x, sec=isec)
+                    exec("sim.neuron.h.setdata_" + self.prefix + "(iseg.x, sec=isec)")
```
mgeplf: I try and not use exec whenever possible.

Does:
`getattr(sim.neuron.h, "setdata_" + self.prefix)(iseg.x, sec=isec)`
work?

examples/l5pc/hocmodel.py
```diff
@@ -37,6 +37,8 @@ def instantiate(self):
         """Instantiate model in simulator"""
 
         self.cell = bglibpy.Cell(self.template, self.morphology)
+
+        self.cell.re_init_rng()
```
wvangeit: Would it be possible to use the bglibpy Simulation() object here ? It would wrap this kind of setup problems.

PR #170, opened 2016-10-20
adding base seed

croessert:
Adding a base seed to mechanism for Stoch channels


bluepyopt/ephys/mechanisms.py
```diff
@@ -121,7 +123,7 @@ def instantiate_determinism(self, deterministic, icell, isec, sim):
                     seg_name = '%s.%.19g' % (short_secname, iseg.x)
                     getattr(sim.neuron.h,
                             "setdata_%s" % self.prefix)(iseg.x, sec=isec)
-                    seed_id1 = icell.gid
+                    seed_id1 = icell.gid+self.base_seed
```
wvangeit: Mmm, we shouldn't set the base_seed this way. Random123 now has https://www.neuron.yale.edu/neuron/static/docs/help/neuron/general/classes/random.html#Random123_globalindex, which will be used by Neurodamus.
Not sure what to do. I could merge this for now, but it's gonna break backward compatibility later ...

croessert: Yes right, I agree. Using globalindex in Random123 is better.

PR #194, opened 2017-02-06
Add int_settings argument to efelfeature

wvangeit:


wvangeit:
@christian-oreilly @croessert @elisabettai The merge of this PR has added an argument to the eFELFeature constructor that allows you to set the integer settings of the eFEL. So you can enable strict interval checking by settings this argument to: int_settings={'strict_stiminterval': True}

wvangeit:
For an example, see here: https://github.com/BlueBrain/BluePyOpt/blob/master/bluepyopt/tests/test_ephys/test_features.py#L83

wvangeit:
Also mentioning  @basjanzandt

PR #212, opened 2017-05-18
Convert to Python3

ralic:
using : find . -name '*.py' | xargs 2to3 -w


bluepyopt/_version.py
```diff
@@ -86,20 +86,20 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False,
             if e.errno == errno.ENOENT:
                 continue
             if verbose:
-                print("unable to run %s" % dispcmd)
+                print(("unable to run %s" % dispcmd))
```
mgeplf: it's better to do it by hand, otherwise things like this show up, IMO

wvangeit:
The BluePyOpt code is already running on Python 3, so no need to run 2to3 on it.

bluepyopt/_version.py
ralic: Thanks , what you said is to the point.

PR #228, opened 2017-10-02
Fix bug where NetCon references are overwritten

mananatee:
Hi there,

I encountered a bug where NetCon objects were overwritten / discarded in case of multiple synapses per 'NrnPointProcessLocation' object.

I also  changed a misleading Debug message that made it seem as if a mechanism is inserted more times than it should.

Cheers,
Lucas.

### DIFF
--- a/bluepyopt/ephys/mechanisms.py
+++ b/bluepyopt/ephys/mechanisms.py
@@ -109,9 +109,9 @@ class NrnMODMechanism(Mechanism, serializer.DictMixin):
                     isec,
                     sim)
 
-            logger.debug(
-                'Inserted %s in %s', self.suffix, [
-                    str(location) for location in self.locations])
+        logger.debug(
+            'Inserted %s in %s', self.suffix, [
+                str(location) for location in self.locations])
 
     def instantiate_determinism(self, deterministic, icell, isec, sim):
         """Instantiate enable/disable determinism"""
@@ -298,9 +298,9 @@ class NrnMODPointProcessMechanism(Mechanism):
             except AttributeError as e:
                 raise AttributeError(str(e) + ': ' + self.suffix)
 
-            logger.debug(
-                'Inserted %s at %s ', self.suffix, [
-                    str(location) for location in self.locations])
+        logger.debug(
+            'Inserted %s at %s ', self.suffix, [
+                str(location) for location in self.locations])
 
     def destroy(self, sim=None):
         """Destroy mechanism instantiation"""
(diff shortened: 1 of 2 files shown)


bluepyopt/ephys/mechanisms.py
```diff
@@ -109,9 +109,9 @@ def instantiate(self, sim=None, icell=None):
                     isec,
                     sim)
 
-            logger.debug(
-                'Inserted %s in %s', self.suffix, [
-                    str(location) for location in self.locations])
+        logger.debug(
```
wvangeit: If this is moved outside of the locations loop, won't it only print a message for the last insert ?

bluepyopt/ephys/stimuli.py
```diff
@@ -140,7 +141,7 @@ def instantiate(self, sim=None, icell=None):
                 netcon = sim.neuron.h.NetCon(netstim, synapse)
                 netcon.weight[0] = self.weight
 
-                self.connections[location.name] = (netcon, netstim)
+                self.connections[location.name].append((netcon, netstim))
```
wvangeit: Ok, that makes sense. Didn't think about the use case in which the same location has multiple netstims.

bluepyopt/ephys/mechanisms.py
wvangeit: ok, you're right. didn't see the for loop inside the string generation

bluepyopt/ephys/stimuli.py
mananatee: Well actually it's a bit weird: I have a NrnModPointProcessMechanism object that represents a synapse model at multiple locations (since it has a list of NrnSeclistCompLocation). Then when you make a NrnPointProcessLocation object for that point process it covers multiple synapses and their location. So when you instantiate it, you want the NetCons to all those synapses to survive.

Before this change, my NetCons tal all except the last synapse would disappear so they wouldn't receive events. Even if they all have the same NetStim.

mananatee: What I mean by weird is that it's a bit counterintuitive: one NrnPointProcessLocation can actually cover multiple synapses at distinct locations.

I know you could make a separate NrnMODPointProcessMechanism for each individual synapse, each with their own NrnPointProcessLocation, but then it's harder to give them all the same parameters.

wvangeit: Yeah, sure, makes sense

PR #251, opened 2018-07-23
Add pip install matplotlib cell in ipython notebook

wizmer:


wvangeit:
Why only matplotlib, and not the other dependencies ?

wizmer:
I thought it was safe to assume the user installed BluePyOpt before opening the notebook. So in this case, matplotlib would be the only one missing.
I can add bluepyopt if you prefer. Up to you.

wizmer:
Actually, to be precise, neuron package could be missing as well. But we can't really install it for the user...

wvangeit:
Yeah, that was actually my point. If we want to this, we probably want to mention all things that have to be installed before using the notebook. But anyway, will merge it for now.
(I only hope nobody will run into trouble if they execute the notebook on a system where they don't have pip install permissions)

PR #265, opened 2019-02-08
Fixed bug 264, empty hall of fame

pineider:
Quick fix for bug 264. In the future we could refactor `DEAPOptimisation.run` to deal with the checkpoint, avoiding inconsistent `population` and `halloffame`.


wvangeit:
@pineider do you fix the test, or should I have a look at it ?

pineider:
> 
> 
> @pineider do you fix the test, or should I have a look at it ?

I haven't yet, I wanted to check them carefully before because I'm not familiar with the code. Could you please have a look at it?

Issue #99, opened 2016-05-26
EFeatureObjective should be abstract


wvangeit:
Make the EFeatureObjective class abstract, and mention it in the documentation

tapaswenipathak:
hi @wvangeit, I created a pr for this long back, would you be reviewing the pr?

wvangeit:
Sorry, we don't need this feature anymore

PR #274, opened 2019-04-27
Made class and method abstract

tapaswenipathak:
Can you please let me know the documentation file?

Fix https://github.com/BlueBrain/BluePyOpt/issues/99.

### DIFF
--- a/bluepyopt/ephys/objectives.py
+++ b/bluepyopt/ephys/objectives.py
@@ -20,12 +20,21 @@ Copyright (c) 2016, EPFL/Blue Brain Project
 """
 
 import bluepyopt
+import abc
 
 
-class EFeatureObjective(bluepyopt.objectives.Objective):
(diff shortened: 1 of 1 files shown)


tapaswenipathak:
Per the ticket the `EFeatureObjective` should be made abstract. If `EFeatureObjective` is made abstract that would make `__init__` and `calculate_feature_scores` as abstract. With `calculate_feature_scores` as abstract and `SingletonObjective` being `single eFeature inside this objective`, if the `calculate_feature_scores` of super class is implemented that is per my syntax wise wrong and doesn't make sense as well. This turned as a research, I am putting up a SO question, I knew I can do w/ MRO but I thought I should make just `EFeatureObjective` no matter what as per the ticket. Let me know if this works!

tapaswenipathak:
hi @wvangeit, who reviews PRs for this repository? I saw your few of your recent commits.

wvangeit:
Sorry, we don't need this feature anymore

PR #284, opened 2019-08-16
tox.ini: Lint for Python 3.7 syntax errors

cclauss:
[__flake8__](http://flake8.pycqa.org) is a superset of pycodestyle that adds many superpowers including the capability to find Python syntax errors and undefined names.

__xrange()__ was removed in Python 3 in favor or a reworked version of __range()__.


wvangeit:
Thanks for pointing this out.
I will use flake8, but will do it in a way so that it runs all the time.
Am fixing the syntax errors too, and will create a separate PR

cclauss:
Closing...  You got the idea.  I am lousy with tox.

PR #285, opened 2019-08-16
Flake8

wvangeit:


wvangeit:
Ok. But thanks for pointing it out. The changes with the syntax fixes are here now:
https://github.com/BlueBrain/BluePyOpt/pull/285

PR #287, opened 2019-09-02
Add timeout to the pool used for parallelization

DrTaDa:
Add a timeout attribute to the CellEvaluator class.
Use the Pebble library instead of the multiprocessing library to instantiate the pool.

### DIFF
--- a/bluepyopt/ephys/evaluators.py
+++ b/bluepyopt/ephys/evaluators.py
@@ -40,9 +40,9 @@ class CellEvaluator(bpopt.evaluators.Evaluator):
             fitness_protocols=None,
             fitness_calculator=None,
             isolate_protocols=None,
-            timeout=None,
             sim=None,
-            use_params_for_seed=False):
+            use_params_for_seed=False,
+            timeout=None):
         """Constructor
 
         Args:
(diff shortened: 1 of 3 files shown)


bluepyopt/ephys/evaluators.py
```diff
@@ -40,6 +40,7 @@ def __init__(
             fitness_protocols=None,
             fitness_calculator=None,
             isolate_protocols=None,
+            timeout=None,
```
wvangeit: Could you move this argument to the end of the list. I'm always a bit afraid people who don't specify labels when calling the function might get in trouble when the order is not maintained

bluepyopt/ephys/protocols.py
```diff
@@ -193,26 +206,30 @@ def _reduce_method(meth):
             import copyreg
             import types
             copyreg.pickle(types.MethodType, _reduce_method)
+            import pebble
+            from concurrent.futures import TimeoutError
 
-            import multiprocessing
+            if timeout is not None:
+                timeout = max(timeout, 0)
```
wvangeit: Wouldn't it be better to raise a ValueError when a negative timeout is given ? This implicit assumption might not be clear

PR #292, opened 2019-10-14
Added stagnation conditions to deapext.

DrTaDa:


bluepyopt/deapext/algorithms.py
```diff
@@ -69,13 +68,22 @@ def _get_offspring(parents, toolbox, cxpb, mutpb):
...
+
+
 def eaAlphaMuPlusLambdaCheckpoint(
         population,
         toolbox,
         mu,
         cxpb,
         mutpb,
         ngen,
+        stagnation=0,
```
wvangeit: Please put these at the end. If people use positional arguments i might break their scripts

bluepyopt/ephys/evaluators.py
```diff
@@ -140,22 +140,32 @@ def run_protocol(
...
         if self.use_params_for_seed:
             sim.random123_globalindex = self.seed_from_param_dict(param_values)
 
-        return protocol.run(
-            self.cell_model if cell_model is None else cell_model,
-            param_values,
-            sim=sim,
-            isolate=isolate,
-            timeout=timeout)
+        try:
```
wvangeit: Please put these (fairly urgently in a separate PR)

bluepyopt/ephys/evaluators.py
```diff
@@ -140,22 +140,32 @@ def run_protocol(
...
+            return protocol.run(
+                self.cell_model if cell_model is None else cell_model,
+                param_values,
+                sim=sim,
+                isolate=isolate,
+                timeout=timeout)
+        except TypeError as e:
+            if "unexpected keyword" in str(e):
```
wvangeit: Mention that this is for backward compatibility in a comment.

bluepyopt/deapext/algorithms.py
```diff
@@ -69,13 +68,22 @@ def _get_offspring(parents, toolbox, cxpb, mutpb):
     return deap.algorithms.varAnd(parents, toolbox, cxpb, mutpb)
 
 
+def _get_median(population, percentage=0.3):
+    '''return the median of the X% fittest individuals'''
+    _ = [ind.fitness.sum for ind in population]
```
wvangeit: Is there a particular reason to use _ ? I'd prefer to use a normal var name.

bluepyopt/deapext/algorithms.py
```diff
@@ -90,6 +98,10 @@ def eaAlphaMuPlusLambdaCheckpoint(
         cxpb(float): Crossover probability
         mutpb(float): Mutation probability
         ngen(int): Total number of generation to run
+        stagnation(int): number of evals after which to stop if the fitness
```
wvangeit: Thinking about this. Wouldn't it be better to implement a 'stop_criterion' class or something. Then that could have different types, like stagnation. And all the parameters of the stop criterion would be hidden inside the class.

DrTaDa: Yes I thought about it. Especially since CMA heavily relies on termination criteria. I will try to come up with something.

On of the issue is that the termination criteria for CMA ask for very different information and thus it will be difficult to write general class.

PR #296, opened 2019-11-29
Modified computation of distance in morphologies.py

danielelinaro:
I have fixed the call to the function neuron.h.distance: as it was, it might have led to the computation of incorrect values of distance between sections.


wvangeit:
Thanks @danielelinaro for contributing. I wonder, based on the documentation, wouldn't setting the origin as sim.neuron.h.distance(0, 0.0, sec=icell.soma[0]) not match better with the old behavior?

danielelinaro:
Hi @wvangeit, in terms of backward compatibility I guess you're right, but to me it seems more logical to use the center of the soma as the origin. That is also the way in which the same function is called in two more files, parameterscalers.py and locations.py, where the origin is the center of the soma.

PR #301, opened 2020-03-04
Stopping criteria

DrTaDa:
Added a StoppingClass class in prevision of the CMA update.
The stopping of IBEA now relies on a MaxNGen class (heritating from StoppingClass) that is triggered when reaching max_ngen.


DrTaDa:
@wvangeit

wvangeit:
Could you add some tests for this new code (should be fully covered)

DrTaDa:
Done !

PR #302, opened 2020-03-31
Fix for loading mechanism library compiled with MPI enabled NEURON

pramodk:
- certain mpi libraries pre-initialize MPI based on certain env
    variables.
  - PMI_RANK is standard SLURM env variable which is used by HPE-MPI
    and can be safely unset

@wvangeit: 
* I don't have full knowledge of bluepyopt toolchain but from my understanding this is where unset should happen.
* I thought it's ok to have this change in bluepyopt so that other users won't face the same problem (external users).

* (Unrelated) : Is it ok to deploy latest release i.e. bluepyopt v1.9.27 and efel 3.0.80 on BB5?

FYI @matz-e @jorblancoa : this is required because even if PMI_RANK is unset in the module, ipengines launched via srun gets new PMI_RANK id.

### DIFF
--- a/bluepyopt/ephys/simulators.py
+++ b/bluepyopt/ephys/simulators.py
@@ -93,7 +93,7 @@ class NrnSimulator(object):
         # MPI before calling MPI_Init (which is undesirable). Unset this
         # variable to avoid issue with loading neuron and mechanism library.
         if 'PMI_RANK' in os.environ:
-            os.environ.pop("PMI_RANK")
+            del os.environ["PMI_RANK"]
 
         import neuron  # NOQA
 
--- a/bluepyopt/ephys/simulators.py
+++ b/bluepyopt/ephys/simulators.py
@@ -89,9 +89,9 @@ class NrnSimulator(object):
             NrnSimulator._nrn_disable_banner()
             self.banner_disabled = True
 
-        # certain mpi libraries (hpe-mpt) use PMI_RANK env variable to initialize
-        # MPI before calling MPI_Init (which is undesirable). Unset this variable
-        # if exist to avoid issue with loading neuron and mechanism library.
+        # certain mpi libraries use PMI_RANK env variable to initialize
+        # MPI before calling MPI_Init (which is undesirable). Unset this
+        # variable to avoid issue with loading neuron and mechanism library.
         if 'PMI_RANK' in os.environ:
             os.environ.pop("PMI_RANK")
 
--- a/bluepyopt/ephys/simulators.py
+++ b/bluepyopt/ephys/simulators.py
@@ -89,6 +89,12 @@ class NrnSimulator(object):
             NrnSimulator._nrn_disable_banner()
             self.banner_disabled = True
 
+        # certain mpi libraries (hpe-mpt) use PMI_RANK env variable to initialize
+        # MPI before calling MPI_Init (which is undesirable). Unset this variable
+        # if exist to avoid issue with loading neuron and mechanism library.
+        if 'PMI_RANK' in os.environ:
+            os.environ.pop("PMI_RANK")
+
         import neuron  # NOQA
 
         return neuron


bluepyopt/ephys/simulators.py
```diff
@@ -89,6 +89,12 @@ def neuron(self):
             NrnSimulator._nrn_disable_banner()
             self.banner_disabled = True
 
+        # certain mpi libraries use PMI_RANK env variable to initialize
+        # MPI before calling MPI_Init (which is undesirable). Unset this
+        # variable to avoid issue with loading neuron and mechanism library.
+        if 'PMI_RANK' in os.environ:
+            os.environ.pop("PMI_RANK")
```
matz-e, line 96:
```suggestion
            del os.environ["PMI_RANK"]
```
? Seems a little more pythonic…

pramodk: True. On [this](https://stackoverflow.com/questions/18178038/how-to-clear-os-environ-value-for-only-one-variable-in-python/18178048) Stackoverflow question I saw people voted more for `pop` and hence used it.

wvangeit:
I could merge this change, at least temporarily. But I'm not convinced this the right solution. 
Can't this be solved in the neuron loading code itself? This is a solution for a very particular situation, but it seems everybody who uses Neuron with hpe mpi / ipyparallel in other code will still be affected?

mgeplf:
I agree with @wvangeit: doing this in every library that uses `NEURON` seems hard.  Also, I think it means that depending on the order of import, the behavior can change.  A simple refactor might mean that `bglibpy` gets loaded after `NEURON`, meaning the MPI code isn't disabled.  Or, alternatively, depending on the order tests are loaded, different behavior is experienced, which is not fun to debug.

wvangeit:
Thinking about it a bit more, I wonder if this can actually also not affect people that don't use HPE MPI. You mentioned that this variable is also used by other MPI implementations? You're sure it won't create problems for these?
For example, I know that @Hjorthmedh is using BluePyOpt to set up neuron models in his network simulator (https://github.com/Hjorthmedh/Snudda). I wonder if unsetting the PMI_RANK can not have consequences in that context ?

pramodk:
Didn't get time to follow up on this, will update here soon.

pramodk:
Here is detailed answer of why this is needed and why I did this PR in BluePyOpt. The issue is complicated and hence the answer. Short summary / recommendation is towards the end.

#### About NEURON
* Last year we went through NEURON codebase to make sure it does right thing (discussion in https://github.com/neuronsimulator/nrn/pull/266). In summary : NEURON doesn't load MPI if user is running stuff serially. So this is not inherent problem in NEURON.
* When user perform `nrnivmodl` on local directory, the created `x86_64/.libs/libnrnmech.so` will have MPI linking. When you import neuron module, this library get loaded and hence the MPI error with HPE-MPI and under SLURM.
* NEURON does support fully dynamic MPI loading i.e. MPI is not linked at all to any library. But this typically not used in cluster environment because : 1) users have to always load "right" MPI library 2) if mod files use any MPI functions (internal to neuron or external), they will get compilation errors. For example, on BB5 we won't be able to compile full neurodamus mod files.
* If users use non-mpi NEURON modules, then issue won't appear. But note that to avoid confusion between "serial" vs "parallel" modules, we have removed all "/serial" modules which I believe is good thing (see [BSD-5](https://bbpteam.epfl.ch/project/issues/browse/BSD-5)). Most of the NEURON installations on clusters, won't have this.

Obviously, we don't want to introduce fixes in other softwares if we can fix stuff in NEURON.

#### About HPE-MPI and BB5

* The peculiar behaviour of HPE-MPI has been discussed with HPE developers (discussion in [HPCTM-1295](https://bbpteam.epfl.ch/project/issues/browse/HPCTM-1295)). In summary : If you are running **inside SLURM** allocation, you can't start HPE-MPI linked application or load mph-linked library without mpirun or srun. This this behaviour doesn't exist with other schedulers like PBS, Torque. So developers don't see this as a priority for HPE-MPI.
* On BB5 when we do salloc, we by default land on allocated node ([original ticket here](HELP-3709)). Which means, our shell session is already running under **srun** and that's why we see `PMI_RANK` variable being set. This is root cause of those all “MPT Error” tickets. (See my tickets on [INFRA-3205](https://bbpteam.epfl.ch/project/issues/browse/INFRA-3205), [HELP-12377](https://bbpteam.epfl.ch/project/issues/browse/HELP-12377)).
*  To allow serial usage of NEURON (NEST, STEPS) in interactive session on BB5, today we unset PMI_RANK variable as part of module. This doesn’t solve the issue with bluepyopt though, more about this below.

#### About bluepyopt / ipengine

* In bluepyopt, the execution workflow is following. Even if we unset PMI_RANK in Step 1. via modules or job script, the ipengines are launched via srun and they will get their own PMI_RANKs.
￼
image

* When code executing on ipengines try to import neuron (or any HPE-MPI linked library), this will result into a MPI error.
*  Hence, we need to unset PMI_RANK variable on the target ipengines.

#### About SLURM

* PMI_RANK is slurm's internal environmental variable.  `ipengine` here is not MPI program and hence we can unset it safely.
* Even if each ipengine will launch another parallel program (e.g. neuron), hypothetical use case, it’s fine because srun internally will set it’s new PMI_RANK
* Along with HPE engineers, I have discussed this issue with SLURM devs. You can see discussion [thread here](https://bugs.schedmd.com/show_bug.cgi?id=8785). 

### My Recommendation (Summary)
* If we want to support bluepyopt with any neuron installation (serial or parallel) on any schedular (slurm or others) with any mpi (hpe-mpi or others), we should unset PMI_RANK.
* This applies to other NEURON consumers if they are using NEURON in setup similar to what I described above. Thee are other alternatives : ask to use serial neuron, build dynamic-mpi enabled neuron etc. but this imposes unnecessary restriction (especially in cluster environments). 
* It’s valid concern that code re-organisation might re-introduce the bug. I will unset this PMI_RANK variable at the beginning/initialisation of bluepyopt. I don’t know the code enough so I did the change at obvious place.
* Our Spack deployment already apply this patch because : 1) I underpants that the new release with such patch might take time 2) we can't have serial/parallel module mess on BB5 and  2) It’s safe and won’t change anything. I take guarantee of solving any issues from this change. 

I believe this answers questions raised in previous comments. If any more details required on this, I will be happy to answer.

mgeplf:
Thanks for the explanation, @pramodk.

> ...  NEURON codebase to make sure it does right thing...NEURON doesn't load MPI if user is running stuff serially.
& 
> When user perform nrnivmodl on local directory, the created x86_64/.libs/libnrnmech.so will have MPI linking.

Can't the fix that was applied to NEURON be applied to the nrnivmodl generator, so that it too 'does the right thing', and only loads MPI if NEURON is doing MPI.  Then the behavior is uniform: NEURON (and its mechs) only use MPI if told to.

>  the ipengines are launched via srun

Could the scripts that srun be modified to 'srun --mpi=none'?  Rather than some side-effects deep in the code.  This makes it more clear what is happening, imo.

> If we want to support bluepyopt with any neuron installation (serial or parallel) on any schedular (slurm or others) with any mpi (hpe-mpi or others), we should unset PMI_RANK.

To me, it looks like this is solving one use case (launch ipengines) w/o considering the impact on other use cases - for instance what happens if you *are* trying to use bluepyopt in an MPI environment - ie what Werner mentioend above about '...using BluePyOpt to set up neuron models in his network simulator'.

wvangeit:
@pramodk, could you please answer @mgeplf 's questions. I agree with his suggestions.

You have to be aware that the fix you're proposing is a 'very' limited fix for this problem. I'm aware of at least two other packages, and some more scripts that make use of exactly the same mechanism as bluepyopt and that will also be affected by this bug. There should be a more thorough solution for this problem.

Regarding the need for you to deploy this fix without going through this pull request. I've already mentioned that I'm very unhappy about this. Your reason is that the 'new release with such patch might take time'.  To me this doesn't make sense at all. The only large delay is from your side when PRs are made in spack. The moment we press the merge button in the bluepyopt repo, their is a pypi package available in 10-20 minutes or so. 
The rule should be that spack deployment isn't used to force fixes (of which you don't necessarily understand all the consequences) are pushed upon the users.

pramodk:
> @pramodk, could you please answer @mgeplf 's questions. 

yeah, don't have enough bandwidth to respond timely here (with all detailes involved). 

-- For Mike:

> Can't the fix that was applied to NEURON be applied to the nrnivmodl generator, so that it too 'does the right thing', and only loads MPI if NEURON is doing MPI. Then the behavior is uniform: NEURON (and its mechs) only use MPI if told to.

You are absolutely right doing uniform behaviour but there are some more details to it. Note that at runtime it's working fine because universally every NEURON user is passing `-mpi` as command line argument. On the other hand, everyone just uses `nrnivmodl`. If we need to do `right` thing then we have to introduce similar option (`-mpi` or `-no-mpi`) to `nrnivmodl`. Will people use it? Technically feasible but this is not going to solve everything (more about this later in `Relevant Details` section).

> Could the scripts that srun be modified to 'srun --mpi=none'? Rather than some side-effects deep in the code. This makes it more clear what is happening, imo.

If a particular HPC system is configured to use newer PMI interface (often for a reason), user overriding this is bad idea IMO and hence I wouldn't suggest this as a solution. (if it works for specific setup and people want to use it, it's fine. This is outside NEURON anyway).

>  for instance what happens if you are trying to use bluepyopt in an MPI environment - ie what Werner mentioend above about '...using BluePyOpt to set up neuron models in his network simulator'.

This works fine. I explained this in `About SLURM` section: 

```
Even if each ipengine will launch another parallel program (e.g. neuron), hypothetical use case,
it’s fine because srun internally will set it’s new PMI_RANK
```

#### Relevant Details

There are some other relevant details and difficult to discuss them here :
* There are 3-4 different ways how MPI enabled NEURON is executed : with/without `-mpi` argument,  using `mpi4py`,  using `NEURON_INIT_MPI`, explicit call to `h.nrnmpi_init()`. The simplest situation is when NEURON is explicitly initialising MPI internally. But this is not always the case how users use NEURON (e.g. mpi4py).
* Dynamic loading has it's own pitfalls : non-standard name of MPI library, newer Python issues on OSX ([comment here](https://github.com/neuronsimulator/nrn/pull/468#issuecomment-619437268))
* On some systems (e.g. Cray HBP Piz-Daint), the default C/C++ compiler on system is also MPI compiler. (This is special case but we use and support users all the time).
* In BBP use case, we want to use same neurodamus cortex/hippocampus modules with bluepyopt. These modules with be MPI compiled by default and have the same issue.

-- For Werner:

>  I've already mentioned that I'm very unhappy about this. Your reason is that the 'new release with such patch might take time'. To me this doesn't make sense at all.

Please understand that everyone has limited resources and trying their best to solve these problems, support everyone and move forward. Thank you.

> You have to be aware that the fix you're proposing is a 'very' limited fix for this problem.

I know this fix has limitations but also note that we are also solving a corner case here and specifically a bug in the HPE-MPI + SLURM. On every other machine, if I have hit this bug I would have simply done or suggested:
* why not just `module swap hpe-mpi another-mpi`? (yes, all other systems have more MPIs than BB5 which has special reasons for using only `hpe-mpi`)
* if you are using neuron serially, why you want to use HPE-MPI compiled NEURON for it? (yes, on BB5 we have reasons 1) don't want to have `neuron/serial` and `neuron/parallel` 2) users want to use `MPI compiled neurodamus` modules without too much confusion.

> Your reason is that the 'new release with such patch might take time'. To me this doesn't make sense at all.
>  I'm aware of at least two other packages, and some more scripts that make use of exactly the same mechanism as bluepyopt and that will also be affected by this bug. There should be a more thorough solution for this problem.

Agree with this. But again, we are trying to solve

wvangeit:
I'm gonna close this. It seems to be fixed in spack for now, and I don't want to risk creating problems on other systems by removing the PMI_RANK var.

pramodk:
Just to update here - with newer software stack deployment and changes in SLURM on BB5 ([HELP-12377](https://bbpteam.epfl.ch/project/issues/browse/HELP-12377)), we won't need this workaround. 🎉

pramodk:
Just for a historical record - With Slurm 21.08 upgrade, we don't need patch described in this PR 😃

wvangeit:
Thx for the info. Where is the party? ;-)

Issue #304, opened 2020-05-03
example thalamocortical-cell does not support python3


orena1:
For example here:
https://github.com/BlueBrain/BluePyOpt/blob/master/examples/thalamocortical-cell/CellEvalSetup/template.py#L85

iteritems is not longer supported in python3

PR #305, opened 2020-05-04
Changed iteritems by items for py3 compatibility

DrTaDa:
See issue [304](https://github.com/BlueBrain/BluePyOpt/issues/304)


wvangeit:
Is that the only change required to make this python 3 compatible? @DrTaDa

DrTaDa:
I checked for xrange, prints, // and iteritems and only found a few more prints.

PR #307, opened 2020-05-06
allow for external modifers of morphologies

arnaudon:
One can now provide a list of functions that will be applied at the end of cell instantiation.
In particular, this allows for custom replace_axon function without creating a new NrnFileMorphology class inherited from the bluepyopt one to only set the AIS.

I also included the slightly more involved set_n_seg, where the user can specify the frequency ( so we can completely get rid of the inheritance mentioned above).


arnaudon:
PR moved to master branch

PR #308, opened 2020-05-06
allow for external modifers of morphologies

arnaudon:
One can now provide a list of functions that will be applied at the end of cell instantiation.
In particular, this allows for custom replace_axon function without creating a new NrnFileMorphology class inherited from the bluepyopt one to only set the AIS.

I also included the slightly more involved set_n_seg, where the user can specify the frequency ( so we can completely get rid of the inheritance mentioned above).


bluepyopt/ephys/morphologies.py
```diff
@@ -50,17 +50,22 @@ def __init__(
...
-                stub ?
-            replace_axon_hoc(str): String replacement for the 'replace_axon'
-            command in hoc  Must include 'proc replace_axon(){ ... }  If None,
-            the default replace_axon is used in any created hoc files
+            do_replace_axon (bool): Does the axon need to be replaced by an AIS
+                stub with default function ?
+            replace_axon_hoc (str): String replacement for the 'replace_axon'
+                command in hoc  Must include 'proc replace_axon(){ ... }
+                If None,the default replace_axon is used
+            do_set_nseg (bool/float): if True, it will use default frequency
```
wvangeit, line 65: I find this naming a bit confusing. When I see this name 'nseg' I'd assume we're setting the nseg to a fixed value, but this is not the case.

bluepyopt/ephys/morphologies.py
```diff
@@ -50,17 +50,22 @@ def __init__(
             do_replace_axon=False,
             do_set_nseg=True,
             comment='',
-            replace_axon_hoc=None):
+            replace_axon_hoc=None,
+            morph_modifiers=None):
```
wvangeit, line 54: i think we're going to have a check in create_hoc of the model to make sure morph_modifiers is None. Because if it's not none, the hoc code conversion will not be valid.

bluepyopt/ephys/morphologies.py
arnaudon: this is the code that is in proj82/singlecell-optimisation, I'm happy to modify it to the better. What would you suggest? replace do_set_nseg -> set_nseg ?

bluepyopt/ephys/morphologies.py
arnaudon: I'm on that, I will push in a minute.

bluepyopt/ephys/morphologies.py
wvangeit: I'd say nseg_frequency, or nseg_div. What do you think?

bluepyopt/ephys/morphologies.py
arnaudon: So I basically allowed the user to add custom hoc files in a list, which are then appended to the replace_axon hoc file. I hope it'll work!

bluepyopt/ephys/morphologies.py
arnaudon: it uses div in the code, but I would better understand frequency

bluepyopt/ephys/morphologies.py
arnaudon: changing that name may also break other codes, that's why I left it as it is

bluepyopt/ephys/morphologies.py
arnaudon: changing that name may also break other codes, that's why I left it as it is

bluepyopt/ephys/morphologies.py
arnaudon: I thought the best was to add another argument to provide a nseg_frequency, which will be =40 by default

bluepyopt/ephys/morphologies.py
arnaudon: I thought the best was to add another argument to provide a nseg_frequency, which will be =40 by default

PR #310, opened 2020-06-03
updated test_eFELFeature to use eFEL v3.1

anilbey:
The difference in the voltage_base feature for this test case is roughly 2e-05.


wvangeit:
I've deleted the caches on travis

Issue #252, opened 2018-07-23
Start/Restart an optimization with a deap population 


anirban6908:
Hi,
I would like to start/restart optimization with a population rather than just specifying a random seed. This would really help if after **_n_** generations I want to change the fitness function and reinitialize the optimization with the population at the **_n_**_th generation.

Thanks

roybens:
Hi i solved this locally by changing line 102 at 
bluepyopt\deapext\algorithms.py to:
 cp = pickle.load(open(cp_filename, "rb"))
just adding a b for the binary worked for me

wvangeit:
@roybens this is for restarting from a checkpoint.
If I'm not mistaken I assume @anirban6908 just wants to be able to provide a list of individuals when starting an optimization ?

roybens:
But wouldn't a checkpoint load do just that? start the optimization from the last population or continue it as it was when the checkpoint was saved. Then you can change the fitness and it should go smoothly. I added objective and continued an optimization from checkpoint it seemed to go well. Regardless I think there is unexpected behavior in loading check point since it misses the binary flag in the load.

anirban6908:
Hi @roybens Right now the checkpoint load would not work if the fitness function is modified e.g., adding a new feature to MOO from previous run to current run. 
@wvangeit  I added this functionality to my fork. If you think this could be useful I can send a PR.
Thanks

wvangeit:
@anirban6908 definitely, feel free to create a PR

wvangeit:
(that was my thinking also, loading from checkpoint puts some requirements on the optimization that saves/loads the checkpoint). Have a pure list of individuals removes these constraints)

wvangeit:
@DrTaDa could you have a look at PR, and merge if it's ok like this.

PR #288, opened 2019-09-05
restart optimization with modified fitness - resolves #252

anirban6908:
Hi @wvangeit ,
Sorry for the delayed PR. Hope this helps.
Thanks

PR #321, opened 2020-08-19
Added feature to start optimisation from a provided population

DrTaDa:
Proposition to solve issue #252


bluepyopt/deapext/optimisations.py
```diff
@@ -258,7 +258,8 @@ def run(self,
             offspring_size=None,
             continue_cp=False,
             cp_filename=None,
-            cp_frequency=1):
+            cp_frequency=1,
+            population=[]):
```
wvangeit, line 262: Better to put None?

DrTaDa: Ok.

bluepyopt/deapext/optimisations.py
```diff
@@ -268,7 +269,32 @@ def run(self,
             offspring_size = self.offspring_size
 
         # Generate the population object
-        pop = self.toolbox.population(n=offspring_size)
+        if population is not None:
+
+            if len(population) != offspring_size:
```
wvangeit, line 274: I think we might have to discuss here what we mean with 'population'. I'd say population = parents + offspring. Don't know how far that makes sense here. Should we maybe consider renaming the field people can specify 'offspring' ?

DrTaDa: Ok !

wvangeit:
@anirban6908 This is a proposal by @DrTaDa to move the population restart feature to the run() call. Can you confirm that this code is what you expect the feature to do? Thanks.

anirban6908:
@wvangeit Sorry for the late reply. Looks good if you want to start the optimization with a new population. 
For the PR #288 I was trying to ensure that if at the end of a checkpointed run (say Stage 1) you want to restart the optimization with a modified fitness function (say Stage 2), the statistics for this run will be written in the same logbook as in Stage 1 and the generation will count up from latest generation in Stage 1 - making it a single optimization run combined of both Stage 1 and Stage 2. 

Correct me if I am wrong but in this PR, for the situation I described the situation above, Stage 1 and Stage 2 will be treated as independent optimizations, will write in separate .pkl files (or even overwrite if checkpoint filename is not changed in Stage 2) - which I wanted to avoid since in my opinion this is the continuation of the same optimization (again for the situation I described).

DrTaDa:
Hello @anirban6908,

wvangeit:
@DrTaDa could you have a look at the merge conflict? thx

DrTaDa:
@wvangeit The conflict is solved. Should we wait for the feedback from @anirban6908 before merging ?

wvangeit:
I'll merge it now, but we can discuss further

anirban6908:
Hi @DrTaDa Apologies for the delay. This looks good to me.

Thanks

wvangeit:
@anirban6908 since this is now addressed in https://github.com/BlueBrain/BluePyOpt/pull/321, are you ok with closing this PR?

anirban6908:
Done.

Issue #134, opened 2016-07-21
cvode_minstep not reset after running protocol


wvangeit:
After a protocol that sets cvode_minstep is run, this variable is not reset to it's original state. This can create problems if other protocols don't use cvode_minstep

tapaswenipathak:
I think this is [fixed](https://github.com/BlueBrain/BluePyOpt/search?utf8=%E2%9C%93&q=cvode_minstep&type=) currently?

tapaswenipathak:
@wvangeit gentle re-ping in case you missed this ping.

tapaswenipathak:
@wvangeit gentle reminder in case you missed this ping.

wvangeit:
@DrTaDa, could you check this?

DrTaDa:
@wvangeit, the same simulator is used by all the protocols since the sim attribute belong to the CellEvaluator, so all the protocols use the same cvode_minstep. 

Does it happen that a protocol itself change the cvode_minstep value ?

wvangeit:
Yes, I think it would be possible that different protocols use different cvode_minstep

PR #322, opened 2020-08-20
Reset cvode_minstep after running protocol

DrTaDa:
Related to issue #134 

The default minstep in NEURON is 0.0, so here I hard-coded the value. 

The original value could also be saved and set back after the simulation, however that would mean having copying a value that is not necessarily the expected one, so it might be cleaner with the hard-coded one.

### DIFF
--- a/bluepyopt/ephys/simulators.py
+++ b/bluepyopt/ephys/simulators.py
@@ -147,7 +147,8 @@ class NrnSimulator(object):
         except Exception as e:
             raise NrnSimulatorException('Neuron simulator error', e)
 
-        self.cvode_minstep = save_minstep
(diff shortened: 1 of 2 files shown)


DrTaDa:
The minstep test would have to be changed to simply check that the cvode minstep is zero before and after the run.

wvangeit:
I'd prefer to save the old value. I don't think we should assume Neuron will always put it at zero. For the test, yeah, I'd just test if it's still the same before and after (but don't assume it's 0.0)

DrTaDa:
NEURON's dt drop to 10**-12 at the beginning and end of the step. I am not sure if it is an expected behavior. Therefore there are two points in the time series that do not respect the minstep.

PR #327, opened 2020-11-16
Custom secnames

DrTaDa:
Add the option of giving custom seclist_names and secarray_names

### DIFF
--- a/bluepyopt/ephys/models.py
+++ b/bluepyopt/ephys/models.py
@@ -63,15 +63,15 @@ class CellModel(Model):
     """Cell model class"""
 
     def __init__(
-            self,
-            name,
-            morph=None,
-            mechs=None,
-            params=None,
-            gid=0,
-            seclist_names=None,
-            secarray_names=None
-            ):
+        self,
+        name,
+        morph=None,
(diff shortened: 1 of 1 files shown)


bluepyopt/ephys/models.py
```diff
@@ -96,10 +105,20 @@ def __init__(
...
+            ]
+        else:
+            self.seclist_names = seclist_names
+
+        if secarray_names is None:
+            self.secarray_names = [
+                'soma', 'dend', 'apic', 'axon', 'myelin'
+            ]
+        else:
+            self.secarray_names = secarray_names
```
arnaudon, line 122: what about a check that at least soma/dend/axon exists? Can you run something without any of these?

arnaudon:
+1

bluepyopt/ephys/models.py
DrTaDa: Good question, the other option is to only pass the custom names and to append them to the minimal list.

wvangeit (approved):

DrTaDa:
@wvangeit , this can be merged.

PR #330, opened 2020-12-02
Added tunable soma origin point for parameter scaler

DrTaDa:
Adds the ability to chose the origin point along the soma (0. to 1.) when using NrnSegmentSomaDistanceScaler.

### DIFF
--- a/bluepyopt/ephys/parameterscalers.py
+++ b/bluepyopt/ephys/parameterscalers.py
@@ -96,7 +96,8 @@ class NrnSegmentSomaDistanceScaler(ParameterScaler, DictMixin):
             name=None,
             distribution=None,
             comment='',
-            dist_param_names=None):
+            dist_param_names=None,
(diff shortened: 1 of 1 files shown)


bluepyopt/ephys/parameterscalers.py
```diff
@@ -110,12 +111,17 @@ def __init__(
...
+                from which to compute the distances. Expressed as a fraction
+                (between 0.0 and 1.0).
         """
 
         super(NrnSegmentSomaDistanceScaler, self).__init__(name, comment)
         self.distribution = distribution
 
         self.dist_param_names = dist_param_names
+        self.soma_ref_location = soma_ref_location
+        assert 0. <= self.soma_ref_location <= 1.
```
wvangeit, line 124: I think it's better to raise an Exception

wvangeit: (ValueError)

PR #336, opened 2020-12-10
Add Location class NrnSecSomaDistanceCompLocation

DrTaDa:
Add class that allows to define a Location based on a section name and a distance from soma.


bluepyopt/ephys/locations.py
```diff
@@ -323,6 +323,99 @@ def __str__(self):
             self.soma_distance, self.seclist_name)
 
 
+class NrnSecSomaDistanceCompLocation(Location, DictMixin):
```
wvangeit, line 326: Would there be a possibility to reuse code of the class above?

DrTaDa: I split the function instantiate from the class NrnSecSomaDistanceCompLocation to be able to reuse the second half of it in the new class.

bluepyopt/ephys/locations.py
```diff
@@ -259,6 +259,7 @@ class NrnSomaDistanceCompLocation(Location, DictMixin):
     """Compartment at distance from soma"""
 
     SERIALIZED_FIELDS = ('name', 'comment', 'soma_distance', 'seclist_name', )
+    max_diam = 0.0
```
wvangeit, line 262: This is a class variable. Are you sure you want to share this between all instances of that class?
What's the function of changing this?

DrTaDa: I do not know why we need the check 
`if comp_diam > max_diam:`

Is there was a reason why it is:
```
max_diam = 0
if comp_diam > max_diam:
```
and not ?
`if comp_diam > 0:`


But indeed class variable might not be the proper way.

wvangeit: Yes, that's true. Not sure where that logic came from originally, it was probably used to avoid zero-diameter sections. We can use if comp_diam > 0

DrTaDa: Changed !

PR #345, opened 2021-01-26
Updated test value in test_metaparameter

AurelienJaquier:
after update in multipoint soma connection in NEURON 8 (see https://github.com/neuronsimulator/nrn/commit/6198541ceaf4de86047ef813ec99f00920aaa3cd).


wvangeit:
So if I understand it correctly the main issue here was that the dendrite got connected to soma(1) in the new version of neuroin, instead of soma(0.5) in the old version. Correct?
(soma(1) is more correct)

AurelienJaquier:
Yes, exactly.

wvangeit:
Ok, was just for the record

PR #346, opened 2021-01-26
Update LICENSE.txt

alex4200:
Update of the LICENSE file according to requirement.


LICENSE.txt
```diff
@@ -1,24 +1,9 @@
 BluePyOpt - Bluebrain Python Optimisation Library
 
-BluePyOpt is licensed under the LGPL, unless noted otherwise, e.g., for external 
-dependencies. See file LGPL.txt for the full license. 
-Examples and test are BSD-licensed.
```
wvangeit, line 5: What happened with special case of examples/tests?

alex4200: What do you mean?

jdcourcol: The examples and tests have a different license (BSD) than the source code (LGPL)
The line stating that is removed with your changes.

alex4200: Then please disregard my changes. I was not aware a repo can have multiple licenses. Not sure how to handle those.

wvangeit: I leave this PR open so you can make more changes? Or do I close it for now?

alex4200: Or maybe you can leave it open.

wvangeit: Ok

wvangeit:
Does this have to stay open ?

Issue #215, opened 2017-06-23
Add a way to record features extracted during optimisations


wvangeit:
At the moment the scores of the parameters are returned to the master process. We should also have a way to e.g. get the raw efeature values out. This would be useful for sensitivity analysis e.g.

elisabettai:
It would be very nice to have this enhancement. For example to analyse how features extracted from good models compare to the experimental ones.

wvangeit:
I will see when I have time to implement this. It won't be trivial, because it involves some changes to the current user-facing API. I'll try to implement in a way so that it doesn't affect existing scripts.

elisabettai:
Ok, I see. Another solution would be to re-run models (>10, potentially hundreds parameter combinations). Maybe this can be efficiently done in parallel (using multiprocessing or ipyparallel), by re-using existing code.

wvangeit:
Yes, for now that might be the best solution. You could use the ipyparallel map function for that. You would only have to pass a function that returns the feature values instead of the scores.
Another option is to subclass response or score calculator classes and let them write out the feature values they see to files.

wvangeit:
@DrTaDa, could you have a look at this. It's an often requested feature. Not urgent though.

DrTaDa:
Does #350 answers entirely to what was requested here ?

PR #350, opened 2021-02-23
Add feature computation functions to CellEvaluator, ObjectivesCalculator and Objectives

DrTaDa:
Add the ability to compute feature values in addition of the feature scores using the CellEvaluator.

Linked to #215


wvangeit:
Could you add some tests for this, thanks.

PR #352, opened 2021-02-24
Adding RTD configuration files

alex4200:


wvangeit:
It's still failing because the BBP license header is not commented out in the requirements_doc file:
https://readthedocs.org/projects/bluepyopt/builds/13096026/

alex4200:
Yes of course, fix pushed

PR #354, opened 2021-03-11
Added terminator to gracefully exit optimisation.

AurelienJaquier:


wvangeit:
You think a test could be added for this?

AurelienJaquier:
If we can play with multiprocessing with Jenkins, I could write a small test to see if the new run_next_gen function gets triggered when terminator is set in another process. I'll have a look into it.

wvangeit:
We don't need this in master?

DrTaDa:
@AurelienJaquier chose this branch because it's the one used in BluePyEModel but it's true that it could be better if we had that into master.

PR #363, opened 2021-04-22
Fix example image path in readme to fix documentation

wvangeit:


wvangeit:
@alex4200 I assume this change is what you intended? I noticed that the readthedocs building was failing because of this path.

PR #368, opened 2021-05-07
Fix Tsodyks Markram example

alberto-antonietti:
One of the two notebooks for the TM fitting (`tsodyksmarkramstp.ipynb`) is not compatible with Python3.

This PR should make the notebook Py3 compatible.


alberto-antonietti:
@wvangeit just to be clear, this solved the notebook-side errors, but not the one involving `sellBEA.py` .

Issue #374, opened 2021-07-29
Make `ipyparallel` and `scoop` optional


mgeplf:
A quick grep of the code suggests that `ipyparallel` and `scoop` could be made into optional dependencies.
This would make the package easier to install, and work around errors from the upstream `scoop`, for instance.

I propose adding an `extras_requires` to the `setup.py`, either called `parallelism` that adds both the libraries to the install, or `scoop` for adding scoop.

Thoughts?

mgeplf:
This is partly in response to:
https://github.com/BlueBrain/spack/pull/1230#issuecomment-888175217

wvangeit:
I don't think ipyparallel should be optional though, most users would be using ipyparallel, and I'd prefer to make loading things as easy as possible for the default user.

wvangeit:
But scoop can indeed be removed

mgeplf:
> I don't think ipyparallel should be optional though, most users would be using ipyparallel

I'm not a huge fan of having deps that aren't used, but ok.

> But scoop can indeed be removed

Removed, or put behind an `pip install bluepyopt[scoop]` flag?

wvangeit:
yes, bluepyopt[scoop] would be good.
@AurelienJaquier or @anilbey could you have a look at this?

mgeplf:
I'll do it, no problem.

anilbey:
Thanks @mgeplf

mgeplf:
Closed by: https://github.com/BlueBrain/BluePyOpt/pull/375

PR #375, opened 2021-07-29
make scoop an in installation extra

mgeplf:
* can do `pip install bluepyop[scoop]`
* in case further extras are to be installed, aded
    `pip install bluepyop[all]`, that would also install scoop

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,11 @@ Copyright (c) 2016-2020, EPFL/Blue Brain Project
 import setuptools
 import versioneer
 
+
+EXTRA_SCOOP = [
+    'scoop>=0.7',
+]
+
 setuptools.setup(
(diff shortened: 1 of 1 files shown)


mgeplf:
this fixes #374

setup.py
```diff
@@ -32,12 +37,16 @@
         'pandas>=0.18',
         'deap',
         'efel>=2.13',
-        'scoop>=0.7',
         'ipyparallel',
         'pickleshare>=0.7.3',
         'Jinja2>=2.8',
         'future',
-	'Pebble>=4.3.10'],
+        'Pebble>=4.3.10'
```
mgeplf, line 44: I just fixed a tab here

AurelienJaquier (approved):
Looks good to me!

PR #377, opened 2021-09-30
Small fix in NrnSomaDistanceCompLocation

AurelienJaquier:
I believe it makes more sense to use max_distance here.

### DIFF
--- a/bluepyopt/ephys/locations.py
+++ b/bluepyopt/ephys/locations.py
@@ -292,7 +292,7 @@ class NrnSomaDistanceCompLocation(Location, DictMixin):
             min_distance = min(start_distance, end_distance)
             max_distance = max(start_distance, end_distance)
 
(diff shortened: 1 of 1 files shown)


wvangeit:
Ok, thanks. 
I saw that the code has: "TODO this definitely has to be unit-tested". Do you know if it's in the tests? Otherwise it might be useful to add some 😄

AurelienJaquier:
I would argue that it is tested.
Although the `find_icomp` function is not directly tested (e.g. with a `test_find_icomp` test function), the `comp_x` value returned by the `instantiate` function is tested [here](https://github.com/BlueBrain/BluePyOpt/blob/master/bluepyopt/tests/test_ephys/test_locations.py#L145)
And the `instantiate` function basically just calls the `find_icomp` function with the appropriate arguments.

PR #385, opened 2022-03-01
Implement LFPy backend

alejoe91:
### LFPy-related changes

- Added LFPy as a backend for simulation
     - [LFPyCellModel](https://github.com/alejoe91/BluePyOpt/blob/merging/bluepyopt/ephys/models.py#L484)
     - [LFPySimulator](https://github.com/alejoe91/BluePyOpt/blob/merging/bluepyopt/ephys/simulators.py#L178)
     - [LFPySquarePulse](https://github.com/alejoe91/BluePyOpt/blob/merging/bluepyopt/ephys/stimuli.py#L341)
     - [TimeLFPResponse](https://github.com/alejoe91/BluePyOpt/blob/merging/bluepyopt/ephys/responses.py#L89)
- Added extracellular efeatures class [extraFELFeature](https://github.com/alejoe91/BluePyOpt/blob/merging/bluepyopt/ephys/efeatures.py#L248) that supports computing features from extracellular "template" (avg action potential). Features include:
    - "peak_to_valley": duration between negative and positive peak
    - "halfwidth": full width alf maximum of negative peak
    - "peak_trough_ratio": ratio between peak and trough
    - "repolarization_slope": slope from the first peak to baseline 
    - "recovery_slope": slop from second peak to baseline
    - "neg_peak_relative": relative amplitude of negative peaks with respect to max channel
    - "pos_peak_relative": relative amplitude of positive peaks with respect to max channel
    - "neg_peak_diff": relative timing of negative peaks with respect to max channel
    - "pos_peak_diff": relative timing of positive peaks with respect to max channel
    - "neg_image": normalized potential at time of negative peak 
    - "pos_image": normalized potential at time of positive peak 

### Other changes

- Add `param_dependancies` in parameters
- Add `**morhp_modifiers_kwargs` to Morphology
- Implement `NrnSegmentSectionDistanceScaler` (base class also for `NrnSegmentSomaDistanceScaler`)


DrTaDa (commented):
The setup.py needs to be updated to include LFPy.
Also we will need tests to maintain the test coverage above 90%.

alejoe91:
> The setup.py needs to be updated to include LFPy. Also we will need tests to maintain the test coverage above 90%.

It was not added on purpose to make the LFPy dependency weak. All LFPy imports are local so one can still run _normal_ BPO without having LFPy installed

DrTaDa:
> > The setup.py needs to be updated to include LFPy. Also we will need tests to maintain the test coverage above 90%.
> 
> It was not added on purpose to make the LFPy dependency weak. All LFPy imports are local so one can still run _normal_ BPO without having LFPy installed

LFP needs to be in an extra require then:
```
extras_require={
        "LFPy": ["LFPy"]
}
```

alejoe91:
yeah that could work!

bluepyopt/tests/test_ephys/test_models.py
```diff
@@ -180,6 +180,23 @@ def test_CellModel_destroy():
     assert 0 == len(sim.neuron.h.CellModel_destroy)
 
 
+@pytest.mark.unit
+def test_lfpycellmodel():
```
wvangeit, line 184: we're gonna need way more unit tests to test all the new functionality above. @AurelienJaquier , maybe something for you?

tox.ini
```diff
@@ -20,6 +20,7 @@ deps =
     flake8
     mock
     neuron-nightly
+    LFPy
```
wvangeit, line 23: Shouldn't we also adapt the setup.py? Maybe we can add an 'extra' that includes LFPy

bluepyopt/parameters.py
```diff
@@ -33,6 +34,9 @@ def __init__(self, name, value=None, frozen=False, bounds=None):
         self._value = value
         self.check_bounds()
         self.frozen = frozen
+        self.param_dependancies = param_dependancies
```
wvangeit, line 37: spelling error

bluepyopt/tests/test_ephys/test_models.py
AurelienJaquier: Sure, I'm on it.

wvangeit:
So it seems LFPy requires mpi4py? Installation seems to fail during the tests. Can we disable mpi4py in LFPy, since I assume we don't need it?

bluepyopt/parameters.py
DrTaDa: > So it seems LFPy requires mpi4py? Installation seems to fail during the tests. Can we disable mpi4py in LFPy, since I assume we don't need it?

Yes, it's only used in the network.py module of LFPy, and we do not use that one.

alejoe91:
> So it seems LFPy requires mpi4py? Installation seems to fail during the tests. Can we disable mpi4py in LFPy, since I assume we don't need it?

Unfortunately it's a requirement: https://github.com/LFPy/LFPy/blob/master/requirements.txt

wvangeit:
Could we ask to get it removed from the hard requirements (or to make an extra that doesn't require it?). I'm a bit reluctant to make mpi4py a requirement for the bpopt tests if it is not required.

alejoe91:
> Could we ask to get it removed from the hard requirements (or to make an extra that doesn't require it?). I'm a bit reluctant to make mpi4py a requirement for the bpopt tests if it is not required.

https://github.com/LFPy/LFPy/issues/429

DrTaDa:
@alejoe91 If you agree, I will rollback the changes to parameters,py and parameterscalers.py. 
We were using them for the Hallermann model but it is not use anymore.
It will also require us to do some changes on the multimodal fitting repo such as removing https://github.com/alejoe91/multimodalfitting/blob/cb49b63d0e76433448b9e4ca6ca825cf74598ffa/multimodalfitting/models/model.py#L305

DrTaDa:
@alejoe91 thinking about it more, it might be useful for someone else some day. I am not sure what should be done.

alejoe91:
I think we still use them for the AIS!

DrTaDa:
It doesn't look like it: https://github.com/alejoe91/multimodalfitting/blob/master/cell_models/hay_ais/parameters.json

alejoe91:
I'll fix the typos

bluepyopt/parameters.py
alejoe91: fixed

bluepyopt/ephys/parameters.py
```diff
@@ -206,19 +216,23 @@ def __init__(
...
-    def instantiate(self, sim=None, icell=None):
+    def instantiate(self, sim=None, icell=None, params=None):
         """Instantiate"""
         if self.value is None:
             raise Exception(
                 'NrnSectionParameter: impossible to instantiate parameter "%s"'
                 ' without value' %
                 self.name)
 
+        _values = {"value": self.value}
```
wvangeit, line 227: Can you be more verbose about what's happening here. Also give the variable a more meaningful name.

Do we really need all this logic in the NrnSectionParameter? It seems to make the class overly complex. Does this logic only apply to NrnSectionParameter for some reason, not to other types of parameters?

bluepyopt/ephys/protocols.py
```diff
@@ -192,7 +193,20 @@ def _run_func(self, cell_model, param_values, sim=None):
...
             cell_model.instantiate(sim=sim)
 
-            self.instantiate(sim=sim, icell=cell_model.icell)
+            if isinstance(cell_model, models.LFPyCellModel):
+                LFPyCell = cell_model.LFPyCell
+            else:
+                LFPyCell = None
+
+            self.instantiate(
+                sim=sim, icell=cell_model.icell, LFPyCell=LFPyCell
```
wvangeit, line 202: I'm not sure I understand the necessity. Can't the LFPyCell be store inside the cell_model object? (I.e. why does this logic have to be here?)

bluepyopt/ephys/protocols.py
```diff
@@ -192,7 +193,20 @@ def _run_func(self, cell_model, param_values, sim=None):
...
+            )
+
+            if hasattr(sim, "electrode"):
```
wvangeit, line 205: This looks very hardcoded. Can't you get something based on the type of the sim or cell_model?

bluepyopt/ephys/protocols.py
```diff
@@ -270,15 +284,23 @@ def _reduce_method(meth):
                                        sim=sim)
         return responses
 
-    def instantiate(self, sim=None, icell=None):
+    def instantiate(self, sim=None, icell=None, LFPyCell=None):
         """Instantiate"""
 
         for stimulus in self.stimuli:
-            stimulus.instantiate(sim=sim, icell=icell)
+            if LFPyCell is not None:
```
wvangeit, line 291: Again here, if LFPyCell would be part of icell all this logic was not necessary

bluepyopt/ephys/recordings.py
```diff
@@ -108,3 +108,70 @@ def __str__(self):
         """String representation"""
 
         return '%s: %s at %s' % (self.name, self.variable, self.location)
+
+
+class LFPRecording(Recording):
+
+    """Electrode sesponse to stimulus"""
```
wvangeit, line 115: typo

bluepyopt/ephys/responses.py
```diff
@@ -84,3 +84,37 @@ def plot(self, axes):
             self.response['voltage'],
             label='%s' %
             self.name)
+
+
+class TimeLFPResponse(Response):
+
+    """Response to stimulus"""
```
wvangeit, line 91: Not obvious what the difference is with the class above

bluepyopt/ephys/simulators.py
```diff
@@ -182,3 +182,161 @@ def __init__(self, message, original):
...
+                step. Default is 0.0.
+            random123_globalindex (int): used to set the global index used by
+                all instances of the Random123 instances of Random
+            mechanisms_directory (str): path to the parent directory of the
+                directory containing the mod files. If the mod files are in
+                "./data/mechanisms", then mechanisms_directory should be
+                "./data/".
+        """
+
+        self.LFPyCellModel = LFPyCellModel
```
wvangeit, line 211: I find it weird that the Simulator has a model inside

bluepyopt/ephys/protocols.py
alejoe91: I agree there is probably a better way to do it. The thing is that one can use the LFPy backend without setting an electrode/probe

bluepyopt/ephys/simulators.py
```diff
@@ -182,3 +182,161 @@ def __init__(self, message, original):
...
+        self.electrode = electrode
```
wvangeit, line 212: Same for electrode, these should be outside of these class

bluepyopt/ephys/protocols.py
```diff
@@ -192,7 +193,20 @@ def _run_func(self, cell_model, param_values, sim=None):
...
+                LFPyCell = cell_model.LFPyCell
+            else:
+                LFPyCell = None
+
+            self.instantiate(
+                sim=sim, icell=cell_model.icell, LFPyCell=LFPyCell
+            )
+
+            if hasattr(sim, "electrode"):
+                if any(["LFP" in rec.name for rec in self.recordings]):
```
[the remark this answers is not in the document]
alejoe91: Yes I think we can compare with the [LFPRecording](https://github.com/alejoe91/BluePyOpt/blob/lfpy/bluepyopt/ephys/recordings.py#L113) class

wvangeit (changes requested):

bluepyopt/ephys/protocols.py
alejoe91: I beliebe `icell` is a NEURON object, so we can't add the `LFPy.Cell` object to it. @wvangeit can you confirm?

wvangeit: Ah ok, sorry, you're right. I thought for a sec icell was the bpopt cell object

bluepyopt/ephys/simulators.py
alejoe91: that is required by LFPy since the `simulate` fucntion is from the `LFPy.Cell` object:  https://github.com/LFPy/LFPy/blob/master/LFPy/cell.py#L992

bluepyopt/ephys/protocols.py
wvangeit: But still, not all stimuli will have LFPyCell as argument, right? So this will break code that doesn't accept this argument

bluepyopt/ephys/stimuli.py
```diff
@@ -327,3 +328,88 @@ def __str__(self):
...
+
+        super(LFPySquarePulse, self).__init__()
+        self.step_amplitude = step_amplitude
+        self.step_delay = step_delay
+        self.step_duration = step_duration
+        self.location = location
+        self.total_duration = total_duration
+        self.iclamp = None
+
+    def instantiate(self, sim=None, icell=None, LFPyCell=None):
```
wvangeit, line 360: This argument should not have uppercases in it, it's a variable, not a classname

bluepyopt/ephys/parameters.py
alejoe91: @darshanmandge do you remember why we implemented it this way?

bluepyopt/ephys/simulators.py
wvangeit: but do you already need the simulate function here? Isn't that something for in the simulate below?

bluepyopt/ephys/responses.py
alejoe91: semantically, the `TimeVoltageResponese` is a single intracellulare voltage traace, the `TimeLFPResponse` is extracellular anc can be multichannel. The implementation is indeed the same (so we could inherit...)

bluepyopt/ephys/protocols.py
alejoe91: Yes. Maybe a try except here? I'm skeptical about hardcoding types. For example, [here] (https://github.com/alejoe91/multimodalfitting/blob/master/multimodalfitting/ecode/ecodestimuli.py) we define externally many LFPy-compatible stimuli... alternatively, each stimulus could have a flag `supports_lfp=False` by defalut at class level and we can check that. What do you think?

bluepyopt/ephys/recordings.py
alejoe91: fixed

bluepyopt/ephys/protocols.py
alejoe91: I'll try to draft a better solution

bluepyopt/ephys/simulators.py
alejoe91: I see, so you suggest moving the `lfpycell` and `electrode` arguments directly to the `run` function.

bluepyopt/ephys/protocols.py
alejoe91: actrually, then we'd move the problem to the `protocol.run()`[https://github.com/alejoe91/BluePyOpt/blob/lfpy/bluepyopt/ephys/protocols.py#L189] function and we would need additional parameters there to be propagated everywhere.

bluepyopt/ephys/simulators.py
alejoe91: actrually, then we'd move the problem to the [protocol.run()](https://github.com/alejoe91/BluePyOpt/blob/lfpy/bluepyopt/ephys/protocols.py#L189) function and we would need additional parameters there to be propagated everywhere.

bluepyopt/ephys/parameters.py
darshanmandge: Not sure but could be related to when there are multiple `param_dependencies`. The `_values` could be values of those parameters.

DrTaDa: @alejoe91, as I said in the other comment we were using them for the Hallermann model but it is not use anymore.

bluepyopt/ephys/responses.py
DrTaDa: Something like ?
```
class TimeLFPResponse(TimeVoltageResponse):

    """Response to stimulus"""

    def __init__(self, name, time=None, lfp=None):
        """Constructor
        Args:
            name (str): name of this object
            time (list of floats): time series
            lfp (list of floats): voltage series
        """

        super(TimeLFPResponse, self).__init__(name, time=time, voltage=None)
        self.response["voltage"] = lfp

    def plot(self, axes):
        raise NotImplementedError

```

bluepyopt/ephys/simulators.py
DrTaDa: Something like:
```

class LFPySimulator(NrnSimulator):

    """LFPy simulator"""

    def __init__(self, dt=None, cvode_active=True, cvode_minstep=None,
                 random123_globalindex=None, mechanisms_directory=None):
        """Constructor

        Args:
            dt (float): the integration time step used by neuron.
            cvode_active (bool): should neuron use the variable time step
                integration method
            cvode_minstep (float): the minimum time step allowed for a cvode
                step. Default is 0.0.
            random123_globalindex (int): used to set the global index used by
                all instances of the Random123 instances of Random
            mechanisms_directory (str): path to the parent directory of the
                directory containing the mod files. If the mod files are in
                "./data/mechanisms", then mechanisms_directory should be
                "./data/".
        """

        super(LFPySimulator, self).__init__(
            dt=dt,
            cvode_active=cvode_active,
            cvode_minstep=cvode_minstep,
            random123_globalindex=random123_globalindex,
            mechanisms_directory=mechanisms_directory
        )

    def run(
            self,
            lfpy_cell,
            lfpy_electrode,
            tstop=None,
            dt=None,
            cvode_active=None,
            random123_globalindex=None):
        """Run protocol"""
        import LFPy
        _ = self.neuron

        lfpy_cell.tstart = 0.0
        lfpy_cell.tstop = tstop

        if dt is not None:
            lfpy_cell.LFPyCell.dt = dt

        if cvode_active and dt is not None:
            raise ValueError(
                'NrnSimulator: Impossible to combine the dt argument when '
                'cvode_active is True in the NrnSimulator run method')

        if cvode_active is None:
            cvode_active = self.cvode_active

        if cvode_active is not None:
            self.cvode_active = cvode_active

        if random123_globalindex is None:
            random123_globalindex = self.random123_globalindex

        if random123_globalindex is not None:
            rng = self.neuron.h.Random()
            rng.Random123_globalindex(random123_globalindex)

        probes = [lfpy_electrode] if lfpy_electrode is not None else None

        sim_params = {
            "probes": probes,
            "rec_vmem": False,
            "rec_imem": False,
            "rec_ipas": False,
            "rec_icap": False,
            "rec_variables": [],
            "variable_dt": self.cvode_active,
            "atol": 0.001,
            "to_memory": True,
            "to_file": False,
            "file_name": None,
        }

        try:
            lfpy_cell.simulate(**sim_params)
        except Exception as e:
            raise LFPySimulatorException("LFPy simulator error", e)

        logger.debug("LFPy simulation finished")

```

bluepyopt/ephys/protocols.py
DrTaDa: We could pass cell instead of icell to the instantiate method

bluepyopt/ephys/protocols.py
DrTaDa: With the changes I proposed to the simulator, it could be:
```
           try:
                if isinstance(sim, sim.LFPySimulator):
                    sim.run(
                        lfpy_cell=cell_model.lfpy_cell,
                        lfpy_electrode=cell_model.electrode,
                        tstop=self.total_duration,
                        cvode_active=self.cvode_active)
                else:
                    sim.run(tstop=self.total_duration, cvode_active=self.cvode_active)
```

DrTaDa:
> in the exp models: https://github.com/alejoe91/multimodalfitting/blob/master/cell_models/cell1_211011_3436/parameters.json#L656-L670

Except if that misunderstood, that still works with vanilla BPO as it does not use a parameter dependency.

alejoe91:
> > in the exp models: https://github.com/alejoe91/multimodalfitting/blob/master/cell_models/cell1_211011_3436/parameters.json#L656-L670
> 
> Maybe I misunderstood, but that still works with vanilla BPO as it does not use a parameter dependency.

Maybe you are right! We just need to test that both the hay and the experimental models un fine.

bluepyopt/ephys/efeatures.py
```diff
@@ -226,3 +244,408 @@ def __str__(self):
...
+    fs = 1.0 / np.mean(np.diff(times))  # kHz
+
+    reference_frames = (peak_times * fs).astype(int)
+
+    if isinstance(snippet_len_ms, (tuple, list, np.ndarray)):
+        snippet_len_before = int(snippet_len_ms[0] * fs)
+        snippet_len_after = int(snippet_len_ms[1] * fs)
+    else:
+        snippet_len_before = int((snippet_len_ms + 1) / 2 * fs)
+        snippet_len_after = int((snippet_len_ms - snippet_len_before) * fs)
```
AurelienJaquier, line 614: I find this implementation when `snippet_len_before` is a scalar strange, and it is not consistent with what is described in the docstring under `ms_cut`. Maybe use this:

`snippet_len_before = snippet_len_after = int(snippet_len_ms * fs)`

It would also be consistent with the case when `ms_cut` is an iterable.

bluepyopt/ephys/extra_features_utils.py
```diff
@@ -0,0 +1,500 @@
...
+        Recovery slope of the waveforms
+    """
+    _, peak_idx = _get_trough_and_peak_idx(waveforms)
+    rslope = np.empty(waveforms.shape[0])
+    rslope[:] = np.nan
+
+    time = np.arange(0, waveforms.shape[1]) * (1 / sampling_frequency)  # in s
+
+    for i in range(waveforms.shape[0]):
+        if peak_idx[i] == 0 or peak_idx[i] == waveforms.shape[1]:
```
anilbey, line 344: this can be reduced to a single comparison with the `in` operator
 
`if peak_idx[i] in [0, waveforms.shape[1]]:`

bluepyopt/ephys/simulators.py
```diff
@@ -182,3 +182,161 @@ def __init__(self, message, original):
...
+
+        if len(hoc_so_list) != 1:
+            warnings.warn(
+                "Unable to find Neuron hoc shared library in %s, "
+                "not disabling banner" % nrnpy_path
+            )
+        else:
+            hoc_so = hoc_so_list[0]
+            nrndll = ctypes.cdll[hoc_so]
+            ctypes.c_int.in_dll(nrndll, "nrn_nobanner_").value = 1
```
anilbey, line 252: This is copied paster from the `NrnSimulator` and not used. 
It can be deleted or the static method can become a function to be used in both simulators.

bluepyopt/ephys/simulators.py
```diff
@@ -182,3 +182,161 @@ def __init__(self, message, original):
...
+            NrnSimulator._nrn_disable_banner()
+            self.banner_disabled = True
+
+        import neuron  # NOQA
+        if self.mechanisms_directory is not None:
+            neuron.load_mechanisms(
+                self.mechanisms_directory, warn_if_already_loaded=False
+            )
+
+        return neuron
```
anilbey, line 268: This code is also from the `NrnSimulator` class.

I think this part of the code can be extracted as a function that only requires the `mechanisms_directory` as a parameter.
That way;
* it can be used in both simulators
* neuron library level function definitions stays out of the object scope
* inheritance will not be necessary between the simulations

```python
        import neuron  # NOQA
        if self.mechanisms_directory is not None:
            neuron.load_mechanisms(
                self.mechanisms_directory, warn_if_already_loaded=False
            )

        return neuron
```

DrTaDa:
@anilbey thank for your comment. We are patching this PR on the side. maybe wait a bit before reviewing the code more.

bluepyopt/ephys/objectives.py
```diff
@@ -85,6 +85,32 @@ def __str__(self):
         """String representation"""
 
         return '( %s )' % self.features[0]
+    
+
+class SingletonWeightObjective(EFeatureObjective):
```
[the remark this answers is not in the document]
anilbey: Btw. this class is not a Singleton. You can instantiate multiple objects from it, which is not allowed in a Singleton.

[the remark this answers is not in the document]
wvangeit: @anilbey , about the singleton. The name singleton here doesn't refer to the python concept of singleton, but rather an objective that only contains a single feature

bluepyopt/ephys/simulators.py
wvangeit: 👍🏻

bluepyopt/ephys/protocols.py
```diff
@@ -270,15 +283,23 @@ def _reduce_method(meth):
...
         return responses
 
-    def instantiate(self, sim=None, icell=None):
+    def instantiate(self, sim=None, cell_model=None):
         """Instantiate"""
 
         for stimulus in self.stimuli:
-            stimulus.instantiate(sim=sim, icell=icell)
+            if isinstance(stimulus, LFPStimulus):
+                stimulus.instantiate(lfpy_cell=cell_model.lfpy_cell)
```
DrTaDa, line 291: Should be sim, lfpy_cell

alejoe91: @AurelienJaquier can you take care of it?

alejoe91: > @AurelienJaquier can you take care of it?

actually I'm on it ;)

espenhgn:
Hi @alejoe91, @wvangeit; as `mpi4py` appeared to be a problem, I've made a PR to LFPy (https://github.com/LFPy/LFPy/pull/433) that removes this package dependency altogether (by using similar MPI communication provided via `neuron.h.ParallelContext()`). Seems to work in my own testing but it should be tested also with this PR before merging.

wvangeit:
There are just some formatting issues at the moment.

wvangeit (approved):

wvangeit:
Should we merge this, @alejoe91 @DrTaDa @AurelienJaquier ?

DrTaDa:
Yes !

PR #387, opened 2022-03-15
Raise exception when the voltage becomes unrealistic during a neuron simulation

DrTaDa:
- Add a check that raises an exception if the voltage gets unrealistic
- Replace h.run by hfadvance in the NrnSimulator

### DIFF
--- a/bluepyopt/ephys/simulators.py
+++ b/bluepyopt/ephys/simulators.py
@@ -112,12 +112,19 @@ class NrnSimulator(object):
 
         return neuron
 
+    def check_voltage_valid(self, icell, voltage_bound=200.):
+        """Raises exception if voltage at soma is out of bound"""
+
+        if icell and abs(icell.soma[0](0.5).v) > voltage_bound:
+            raise NrnSimulatorException('Membrane potential is out of bound')
+
     def run(
             self,
             tstop=None,
             dt=None,
             cvode_active=None,
-            random123_globalindex=None):
+            random123_globalindex=None,
+            icell=None):
         """Run protocol"""
 
         self.neuron.h.tstop = tstop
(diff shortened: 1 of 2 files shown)


bluepyopt/ephys/simulators.py
```diff
@@ -112,12 +112,19 @@ def neuron(self):
 
         return neuron
 
+    def check_voltage_valid(self, icell, voltage_bound=200.):
```
wvangeit, line 115: I find this dangerous though. Ideally NEURON should detect if it gets values that it can't handle. Do we know what the effect this has on optimizations? I wouldn't be surprised if a lot of initial solutions might be out of bounds here.

bluepyopt/ephys/simulators.py
```diff
@@ -163,7 +170,10 @@ def run(
             rng.Random123_globalindex(random123_globalindex)
 
         try:
-            self.neuron.h.run()
+            self.neuron.h.finitialize()
```
wvangeit, line 173: Are we sure this 'exactly' replicates the 'run()' behavior of Neuron? We don't want to change results if not necessary.

bluepyopt/ephys/simulators.py
```diff
@@ -163,7 +170,10 @@ def run(
...
+            while self.neuron.h.t < self.neuron.h.tstop:
+                self.neuron.h.fadvance()
+                self.check_voltage_valid(icell)
```
wvangeit, line 176: We also need to check how this affects the simulation time.

wvangeit (changes requested):

bluepyopt/ephys/simulators.py
DrTaDa: Ok,I will open an issue in Neuron

wvangeit: Any news?

DrTaDa: The engineer who was looking into it is just back from vacation. He will resume working on it.

PR #388, opened 2022-03-15
increase max_ngen in CMA

arnaudon:


bluepyopt/deapext/optimisationsCMA.py
```diff
@@ -267,6 +267,9 @@ def run(
             numpy.random.set_state(cp["np_rndstate"])
             CMA_es = cp["CMA_es"]
             CMA_es.map_function = self.map_function
+            # here is to resume optimisation with larger number of generations
+            CMA_es.active = True
+            CMA_es.stopping_conditions[0].max_ngen = max_ngen
```
wvangeit, line 272: Where else is this stopping_conditions used/set? I'm not sure I understand why this only happens when continuing from a cp.

arnaudon: in cma_creator l278 below the max_ngen is set to this list of criterions, so I just update it, as I think it reads the previous values from the optimisation pickle. It also seems to read the active flag, which was set to False as the optimisation had completed.

arnaudon:
And no idea what is happening with the test haha

bluepyopt/deapext/optimisationsCMA.py
wvangeit: About the max_ngen. But shouldn't we use the value from the checkpoint? It seems to me now we might be overriding it with the default value of max_ngen, which is 0.
@DrTaDa about this default value, shouldn't this be 'None' or so?

wvangeit: The same for the active field, shouldn't this also be coming from the checkpoint?

DrTaDa: If we were to implement this suggestion, the optimisation would keep going despite having reached the max_ngen which would go against the purpose of the checkpoint.

@wvangeit yeah, None might be better than 0. Note that if max_ngen is 0, it is set to:
```max_ngen = 100 + 50 * (self.problem_size + 3) ** 2 / numpy.sqrt(self.lambda_)```

wvangeit: Ok, so we should mirror the active value from the checkpoint.

About the None, I think it's better to change the default value to None then. I find it a bit counterintuitive that a value of '0' results in a value of max_ngen that is actually not zero.

arnaudon: ```If we were to implement this suggestion, the optimisation would keep going despite having reached the max_ngen which would go against the purpose of the checkpoint.```
this is the purpose of this change: to resume optimisation from a checkpoint that terminated because it reached max_ngen, but to keep optimising a bit further. I see this can be a usecase, no? I set to 1000, but from the convergence plot, I think 1500 would be better, but I don't want to loose the first 1000 points.

arnaudon: to me, a checkpoint is what it says: state of the result at a point in time, it is for the user to decide if they want to keep running the code or not (and they can change their mind along the way)

wvangeit: ok, but that maybe a check should first be done to see if the value is not None. Because as it is now, when the user doesn't specify the value, it will be set to 0 again.

PR #395, opened 2022-04-29
location along trunk

arnaudon:
new location to consider distance from soma along a path selected using a terminal section furthest away into a direction (in euclidean distance), or radial distance.

### DIFF
--- a/bluepyopt/ephys/locations.py
+++ b/bluepyopt/ephys/locations.py
@@ -450,9 +450,9 @@ class NrnTrunkSomaDistanceCompLocation(NrnSecSomaDistanceCompLocation):
             ]
         )
         if self.direction == 'radial':
-            self.sec_index = np.argmax(np.linalg.norm(points, axis=1))
+            self.sec_index = int(np.argmax(np.linalg.norm(points, axis=1)))
         else:
-            self.sec_index = np.argmax(points.dot(self.direction))
+            self.sec_index = int(np.argmax(points.dot(self.direction)))
 
     def instantiate(self, sim=None, icell=None):
         """ """


wvangeit:
Can you fix the tests?

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,35 @@ def instantiate(self, sim=None, icell=None):
...
 
 
+class NrnTrunkSomaDistanceCompLocation(NrnSecSomaDistanceCompLocation):
+    """Location at a distance from soma along a main direction.
+
+    This is most useful to follow the trunk of an apical dendrite
+    without knowing the apical point, but only that apical trunk goes along y.
+    """
+
+    def set_sec_index(self, icell=None, direction=[0.0, 1.0, 0.0]):
```
wvangeit, line 403: Lists should normally not be used as default values (see e.g. https://docs.python-guide.org/writing/gotchas/). You could set it to None, and immediately assign it to this value on the first line of the method under a 'if direction is None' statement

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,35 @@ def instantiate(self, sim=None, icell=None):
...
+        """Search for the point furthest away along given direction."""
```
wvangeit, line 404: Could you add a little doc about the direction argument.

bluepyopt/ephys/locations.py
wvangeit, line 403: Maybe I'm missing something, but isn't one supposed to specify the distance?

arnaudon: ah yes, I keep forgetting that, but in some codes, lint complains about it, then I don't forget

bluepyopt/ephys/locations.py
arnaudon: I added more doc!

bluepyopt/ephys/locations.py
arnaudon: ah this is just to find the sec index of the last section, then we use normal bpopt location which picks a section at a given location between soma and that section. Basically, instead of apical section id, we pick the furthest away, so we don't have to find the apical point, it's the accurate enough (if an apical trunk is <400, the dend2 feature we have will be in the tuft, but won't completely crash, or if apical point is spoted near soma as we have an early bifucration cell such as L5_TPC:B, then we'll pick one of the secondary apicla trunks, ...)

wvangeit: Sorry, I confused set_sec_index with the constructor. So the original constructor is still there, all fine then

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,44 @@ def instantiate(self, sim=None, icell=None):
         return self.find_icomp(sim, branches)
 
 
+class NrnTrunkSomaDistanceCompLocation(NrnSecSomaDistanceCompLocation):
+    """Location at a distance from soma along a main direction.
+
+    We search for the section that is the furthest away from some along a direction,
+    and pick a location at a given distance from soma along the path to that section.
```
arnaudon, line 400:
```suggestion
    We search for the section that is the furthest away from some along 
    a direction, and pick a location at a given distance from soma along 
    the path to that section.
```

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,44 @@ def instantiate(self, sim=None, icell=None):
...
+
+    This is most useful to follow the trunk of an apical dendrite
+    without knowing the apical point, but only that apical trunk goes along y.
+    """
+
+    def set_sec_index(self, icell=None, direction=None):
+        """Search for the point furthest away along given direction.
+
+        Args:
+            direction (list of 3 elements): 3d vector representing the direction
```
arnaudon, line 410:
```suggestion
            direction (list of 3 elements): 3d vector representing direction
```

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,46 @@ def instantiate(self, sim=None, icell=None):
         return self.find_icomp(sim, branches)
 
 
+class NrnTrunkSomaDistanceCompLocation(NrnSecSomaDistanceCompLocation):
+    """Location at a distance from soma along a main direction.
+
+    We search for the section that is the furthest away from some along 
+    a direction, and pick a location at a given distance from soma along 
+    the path to that section.
+    the path to that section.
```
arnaudon, line 402:
```suggestion
```

bluepyopt/ephys/locations.py
wvangeit: But you're using euclidian distances, is that really what you want? Wouldn't one expect to use the path distance? (or was that just too difficult to implement?).
Actually, are you even using euclidian distances? Since you're allowing the user to specific any direction, shouldn't you have some norm equation in this?

arnaudon:
ah, my commit appeared, but it took a while.... sorry for the noise

bluepyopt/ephys/locations.py
arnaudon: I just project on the vector direction, and take the largest value, I could add radial distance, or path distance, but I'm not sure it would be so useful. The direction is for example for horizontal or inverse PC cells.

wvangeit: just to be sure, will this if one specifies a diagonal direction, or if the soma is not centered at 0,0,0?
I just want to make sure this is generally applicable if people specify a general direction.

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,73 @@ def instantiate(self, sim=None, icell=None):
...
+        super(NrnTrunkSomaDistanceCompLocation, self).__init__(
+            name,
+            soma_distance=soma_distance,
+            sec_index=sec_index,
+            sec_name=sec_name,
+            comment=comment
+        )
+
+        if direction is None:
+            direction = [0.0, 1.0, 0.0]
```
wvangeit, line 436: Could you mention this default value in the docstring with the arguments, this is not trivial.

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,73 @@ def instantiate(self, sim=None, icell=None):
         return self.find_icomp(sim, branches)
 
 
+class NrnTrunkSomaDistanceCompLocation(NrnSecSomaDistanceCompLocation):
```
wvangeit, line 396: We also need a test for this class

arnaudon: the test for parent class `TestNrnSecSomaDistanceCompLocation` is also missing, but I cannot make it work, the mock `Cell` object does not have `soma[0]` as parent of first section, and I'm not sure how to fix this. If you have any suggestions, I'll give it a try!

arnaudon: test added, thank aurelien!

arnaudon:
tox is happy in local on bb5 with python3.9, I'm not sure what is happening here, I'm not familiar enough with this code's CI, etc... @AurelienJaquier , you seem to be the one able to fix that, thank you!

AurelienJaquier:
I tried this locally, and it worked fine for python3.7, but failed, as it did in the CI, for python3.6. Proposed solution: dropping python 3.6 support :P

AurelienJaquier:
I think it is due to this change introduced in python 3.7: https://docs.python.org/3/whatsnew/3.7.html#itertools

bluepyopt/ephys/locations.py
```diff
@@ -392,6 +393,74 @@ def instantiate(self, sim=None, icell=None):
...
+                    section.y3d(section.n3d() - 1),
+                    section.z3d(section.n3d() - 1),
+                ]
+                for section in getattr(icell, self.seclist_name)
+            ]
+        )
+        if self.direction == 'radial':
+            self.sec_index = np.argmax(np.linalg.norm(points, axis=1))
+        else:
+            self.sec_index = np.argmax(points.dot(self.direction))
```
AurelienJaquier, line 455: I recommend using `self.sec_index = int(np.argmax(points.dot(self.direction)))` for compatibility with python 3.6 when `sec_index` is later used in `itertools.islice()`

AurelienJaquier:
The problem comes from, in python 3.6, itertools.islice expect a strict int, and not a numpy.int. The changes I propose makes the new class compatible with python 3.6

arnaudon:
Yay! Thank you @AurelienJaquier , your' the best!

wvangeit (approved):

PR #397, opened 2022-05-11
Stochastic mech features

AurelienJaquier:
- new method that either set to deterministic, or set back to stochastic (if stochastic mechanism)
- always set stoch data, even when used in deterministic mode. Reason: some neuron function (e.g. psection) will look up the mechanism data, and if it points to nothing, it crashes

### DIFF
--- a/bluepyopt/ephys/mechanisms.py
+++ b/bluepyopt/ephys/mechanisms.py
@@ -85,6 +85,9 @@ class NrnMODMechanism(Mechanism, serializer.DictMixin):
         self.preloaded = preloaded
         self.cell_model = None
         self.deterministic = deterministic
+        # If deterministic is False, it might get changed to True.
+        # Use this variable to change it back to False when needed.
+        self.always_deterministic = deterministic
 
         if prefix is not None and suffix is not None:
             raise TypeError('NrnMODMechanism: it is not allowed to set both '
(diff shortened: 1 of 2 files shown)


DrTaDa:
@wvangeit this is something that was implemented for BluePyEModel but we thought it might be better in BluePyOpt

bluepyopt/ephys/mechanisms.py
```diff
@@ -85,6 +85,9 @@ def __init__(
         self.preloaded = preloaded
         self.cell_model = None
         self.deterministic = deterministic
+        # If deterministic is False, it might get changed to True.
```
wvangeit, line 88: I think we need a bit better description here. It's a bit weird code, so I think the use case should be documented well

bluepyopt/ephys/mechanisms.py
```diff
@@ -143,6 +147,13 @@ def instantiate_determinism(self, deterministic, icell, isec, sim):
                     'Stoch channel, not %s' %
                     self.suffix)
 
+    def set_stochasticity(self, allow_stochasticity):
+        """Set stochasticity if allowed, else force determinism"""
+        if allow_stochasticity and not self.always_deterministic:
+            self.deterministic = False
+        else:
+            self.deterministic = True
```
wvangeit, line 155: Would there be a possibility of raising an exception here. If I call set_stochastic on something, I assume the object will be stochastic afterwards, unless if it throws an exception.

AurelienJaquier: The function is more of 'whether this should be stochastic or not', than 'set this to stochastic'. The idea is that, for the pre-protocols, we want everything to be deterministic, so we use this method with allow_stochasticity=False, which sets everything to deterministic. Once the preprotocols are done, we want to set back the stochasticity of each mechanism to its original state (True/False), and so we call this function with allow_stochasticity=True. Note that even when stochasticity is allowed, not every mech will be set to non-deterministic.

bluepyopt/ephys/mechanisms.py
AurelienJaquier: 'One might want to temporarily set a stochastic mechanism to deterministic to run a few protocols, and then turn it back to stochastic mode to run some other protocols. The always_deterministic attribute remembers the mechanisms that were set to be deterministic or stochastic when created. If one wants to turn the stochasticity back on, after temporarily forcing the mechanisms to be deterministic, one can set the deterministic attribute to the value of always_deterministic using the self.set_stochasticity(True) method'

Would a comment like this be acceptable?

bluepyopt/ephys/mechanisms.py
DrTaDa: (pre-protocols are protocols such as the ones to search for RMP, Rin, rheobase ..)

DrTaDa:
@AurelienJaquier it makes me think, maybe all of the stochastic logic could be brought to BluePyOpt. We should have a look.

AurelienJaquier:
@DrTaDa Do you mean the logic in the run() method of BPEM_Protocol? Or were you thinking of something else?

wvangeit:
Should the pre_protocol change the value to deterministic, run it, and set it back to the old value?

DrTaDa:
> @DrTaDa Do you mean the logic in the run() method of BPEM_Protocol? Or were you thinking of something else?

@AurelienJaquier yes, this part.

AurelienJaquier:
Ok, I'll try both your ideas.

DrTaDa:
Looks good to me

wvangeit (approved):

PR #398, opened 2022-06-01
Activate coverage during tests

AurelienJaquier:


Makefile
```diff
@@ -67,8 +67,6 @@ clean:
 	rm -rf examples/l5pc/L5PC.py
 	rm -rf examples/l5pc/x86_64
 	rm -rf examples/stochkv/x86_64
-	rm -rf .coverage
```
wvangeit, line 70: I'm not sure I understand this. The 'make clean' is called before all the tests, right? Will this not have the danger of merging coverage of previous runs of the tests?

wvangeit: But is 'make clean' called between the unit/functional tests? Doesn't seem like it from the tox.ini file.

AurelienJaquier: It is. When you look at the logs of the tests, you have the `rm -rf ...` lines from the make clean before each test.

wvangeit: Also, let's move the two 'functional:' lines together at the end, so that the unit tests run before the mod files are compiled

wvangeit (approved):

AurelienJaquier:
wait don't merge yet

wvangeit:
ok

AurelienJaquier:
Alright, I think I'm good on my side

PR #393, opened 2022-04-07
Arbor cable cell exporter and backend

lukasgd:
I'm working on an [Arbor cable cell](https://docs.arbor-sim.org/en/stable/python/cable_cell.html) exporter for BluePyOpt  optimized cell models (related to https://github.com/arbor-sim/arbor/issues/1839) and looking for feedback here.

I've added a `create_acc` module analogous to `create_hoc` in `bluepyopt.ephys`. It works with the `simplecell` and `l5pc` examples. Using `generate_acc.py` in the corresponding example folder, you can output a cell with optimized parameters to 3 files (plus morphology) - a JSON file for metainformation that references a morphology file as well as two ACC files - one for the decor and one for the label dictionary. The output of running
```
$ BluePyOpt/examples/simplecell/generate_acc.py -o <output-dir>
```
is
```
$ tree <output-dir>
<output-dir>
├── simple_cell_decor.acc
├── simple_cell.json
├── simple_cell_label_dict.acc
└── simple.swc

0 directories, 4 files
```
with file contents
* simple_cell.json
```json
{ 
  "cell_model_name": "simple_cell",
  "produced_by": "Created by BluePyOpt(1.12.5) at 2022-04-06 18:17:54.383572", 
  "morphology": "simple.swc",
  "label_dict": "simple_cell_label_dict.acc",
  "decor": "simple_cell_decor.acc"
}
```
* label_dict.acc
```
(arbor-component
  (meta-data (version "0.1-dev"))
  (label-dict 
    (region-def "all" (all)) 
    (region-def "apic" (tag 4)) 
    (region-def "axon" (tag 2)) 
    (region-def "dend" (tag 3)) 
    (region-def "soma" (tag 1)) ))
```
* decor.acc
```
(arbor-component
  (meta-data (version "0.1-dev"))
  (decor
    (paint (region "soma") (membrane-capacitance 1))
    (paint (region "soma") (density (mechanism "hh" ("gnabar" 0.10299326453483033) ("gkbar" 0.027124836082684685))))))
```
For the `l5pc` the output is more comprehensive (obtained analogously with `BluePyOpt/examples/l5pc/generate_acc.py`). 
```
$ tree <output-dir>
<output-dir>
├── C060114A7.asc
├── l5pc_decor.acc
├── l5pc.json
└── l5pc_label_dict.acc
```
with an analogous JSON and the same label_dict as above, but different decor,
```
(arbor-component
  (meta-data (version "0.1-dev"))
  (decor
[...]

PR #406, opened 2022-06-23
Fix tests CMA

DrTaDa:


DrTaDa:
@wvangeit to fix the issue with #393

tox.ini
```diff
@@ -19,7 +19,7 @@ deps =
     coverage
     flake8
     mock
-    neuron-nightly
+    neuron
```
wvangeit, line 22: I'd prefer keep on testing on nightly though, to find issues as early as possible

DrTaDa: I think we just found an issue then :D

DrTaDa: https://github.com/BlueBrain/BluePyOpt/actions/runs/2549275161

DrTaDa: Is it something we have to report to the nrn github ?

wvangeit: Probably best. This is an error generator by a neuron header  file, right?

DrTaDa: The issue is coming from this line: https://github.com/BlueBrain/BluePyOpt/blob/cc5448730d73311150421a4bec0930315d39cf9b/examples/stochkv/mechanisms/StochKv3.mod#L110

DrTaDa: Ah Pramod answered: https://github.com/neuronsimulator/nrn/issues/1875

pramodk: As mentioned in the [BBP internal slack channel](https://bluebrainproject.slack.com/archives/C02UDSR3AAY/p1655995850804189), assuming this StochKv3.mod from BBP, I would copy the updated version from BBP models (e.g. [here](https://bbpgitlab.epfl.ch/hpc/sim/models/neocortex/-/blob/main/mod/v6/StochKv3.mod))

Would be happy to help if anything else is required.

wvangeit:
I see that there are quite a few changes in examples/l5pc/mechanisms/ProbAMPANMDA_EMS.mod and GABA. I think we don't use these files? Maybe it's better to just remove these from the repo

DrTaDa:
We use both in the examples/stochkv

wvangeit:
Ok, but not the synapse files i assume, right?

bluepyopt/tests/test_deapext/test_optimisationsCMA.py
```diff
@@ -57,4 +58,5 @@ def test_optimisationsCMA_MO_run():
 
     assert abs(log.select("avg")[-1] - 40.) < 1e-4
     assert abs(log.select("std")[-1] - 16.32993) < 1e-4
-    assert pop[0] == [0.09601241274168831, 0.024646650865379722]
+    assert abs(pop[0][0] - 0.09601241274168831) < 1e-6
```
mgeplf, line 61: FYI: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx

wvangeit (approved):

Issue #424, opened 2022-11-20
isolate_protocols=True causing NEURON to reset dt


wthun:
When passing a simulator with a fixed, non-standard time step to ephys.evaluators.CellEvaluator, I receive:

```
Exception: NrnSimulator: Some process has changed the time step dt of Neuron since the creation of this NrnSimulator object. Not sure this is intended:
current dt: 0.025
init dt: 0.1
```

This only happens when setting isolate_protocols to True, so I guess this happens due to the dt value not persisting when NEURON is imported onto new threads (i.e. around line 252 in protocols.py)? 

I see the same behaviour in a [minimal example](https://github.com/BlueBrain/BluePyOpt/files/10050756/bpopt.timestep.bug.zip).

DrTaDa:
Hello @wthun , 

Thank you for opening this issue.

The error is expected in this case. It happens because you ask both for cvode=True and a specific time step dt of 0.1. These two requirements are incompatible since cvode means varying time step. Therefore if  you wish to use cvode you shouldn't pass a dt when initializing the simulator object.

The error message is not very explicit. We will work on it.

wthun:
Hello,

I seem to get the same error when disabling cvode. There is an example attached in the original post, where passing
`nrn = ephys.simulators.NrnSimulator(dt=dt, cvode_active=False)` and then passing that simulator to the evaluator object results in the same error.

DrTaDa:
Indeed. I am sorry I misunderstood your example. We will fix that.

wthun:
No worries and thanks for the help!

AurelienJaquier:
Hi!
I can run your example without any error in linux, but the error shows up when I run under macos. Do you confirm @wthun that you also encountered this error on macos?

It looks like this is out of the scope of BluePyOpt, and should be solved at the NEURON level. I'll open an issue there first, and see if we have to change BluePyOpt once they have acted on it.

Thanks for raising this issue @wthun

wthun:
Yes, but I originally encountered this on Linux. I receive the same error If I run the same example using Ipyparallel and passing lview.map_sync to the optimiser (([like this](https://github.com/BlueBrain/BluePyOpt/files/10061060/bug_parallel.py.zip)). In that case, isolate_protocols doesn't seem to help either.

AurelienJaquier:
Alright, the latest version of BluePyOpt should work fine with custom `dt`, both when `isolate_protocols=True` on macos, and when using ipyparallel.

[...]

PR #425, opened 2022-11-22
fork subprocess and set neuron dt before ipyp process

AurelienJaquier:
Should solve Issue #424 

- for macos and py3.8+, subprocesses are spawn by default. This reset neuron.h.dt in isolate mode due to pebble. Using a context with fork (instead of spawn) solves this.
- neuron dt was also reset in ipyparallel processes. This creates a function to set the neuron variables (loading stdrun.hoc, setting dt, cvode) and calls it in toolbox.map().

### DIFF
--- a/bluepyopt/ephys/evaluators.py
+++ b/bluepyopt/ephys/evaluators.py
@@ -216,6 +216,17 @@ class CellEvaluator(bpopt.evaluators.Evaluator):
 
         return self.objective_list(obj_dict)
 
+    def set_neuron_variables_and_evaluate_with_lists(
+        self, param_list=None, target='scores'
+    ):
+        """Set neuron variables and run evaluation with lists.
+        
+        Setting the neuron variables is necessary when using ipyparallel,
+        since the new subprocesses have pristine neuron.
+        """
+        # self.sim.set_neuron_variables()
+        return self.evaluate_with_lists(param_list=param_list, target=target)
+
     def evaluate(self, param_list=None, target='scores'):
         """Run evaluation with lists as input and outputs"""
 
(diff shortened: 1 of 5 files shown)


bluepyopt/ephys/evaluators.py
```diff
@@ -216,6 +216,17 @@ def evaluate_with_lists(self, param_list=None, target='scores'):
 
         return self.objective_list(obj_dict)
 
+    def set_neuron_variables_and_evaluate_with_lists(
+        self, param_list=None, target='scores'
+    ):
+        """Set neuron variables and run evaluation with lists.
+
+        Setting the neuron variables is necessary when using ipyparallel,
+        since the new subprocesses have pristine neuron.
```
DrTaDa, line 225: maybe NEURON to destinguish from the neuron model ?

bluepyopt/ephys/protocols.py
```diff
@@ -245,12 +245,20 @@ def _reduce_method(meth):
             copyreg.pickle(types.MethodType, _reduce_method)
             import pebble
             from concurrent.futures import TimeoutError
+            import multiprocessing
+
+            # Default context for python>=3.8 on macos is spawn.
+            # Spwan context would reset neuron properties, such as dt.
+            multiprocessing_context = multiprocessing.get_context('fork')
```
DrTaDa, line 252: If fork does not reset the NEURON properties, why do we need this and to call set_neuron_variables?

bluepyopt/ephys/evaluators.py
AurelienJaquier: Done in latest commit

DrTaDa (approved):

AurelienJaquier:
I don't have the rights to merge. Could you do it if you have them @DrTaDa ?

PR #429, opened 2022-12-09
Add publication on TestPyPi test

AurelienJaquier:


AurelienJaquier:
do not merge yet, I'm using this PR to see if the new test behaves as expected

AurelienJaquier:
The pipeline failed because it could not find the github token. It might be due to the fact that workflows can access token only once approval has been given.
Could you approve this PR so that I can test if this makes the workflow use the token @DrTaDa ?

DrTaDa (approved):

wvangeit:
I've updated the token to be sure. Am rerunning the tests

AurelienJaquier:
It still did not pass. It might be because I am opening this PR from a fork repo. Using pull_request_target in github actions might be the trick to solve this.

wvangeit:
I guess it makes sense this doesn't work. If external PRs would use these secret, people could just echo them.

wvangeit:
So not sure there is a solution to test a pypi deployment

AurelienJaquier:
Yeah, but pull_request_target should prevent people from steeling the secret: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

AurelienJaquier:
It looks like the actions are stuck for whatever reason. I'll close the PR and re-open it to retrigger the actions

AurelienJaquier:
I think pull_request_target does not  trigger the workflow properly. I will drop this PR, it does not look like there is an easy way to enable the use of secrets in a secure way on forked pull requests.

PR #442, opened 2023-01-19
add a public generate_parameters function

anilbey:
EModelRunner and e-model-packages have been depending on the private `_generate_parameters`.

It's recent signature change cause failures in the dependent packages.

This PR adds a public function so that the dependent packages won't have to worry about bluepyopt internals.


AurelienJaquier (approved):
Nice, thank you!

PR #446, opened 2023-02-21
drop python3.7 support

anilbey:
Py3.7's end of life is in 4 months. 
https://endoflife.date/python

Our NEURON dependency already dropped its support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,8 @@ Copyright (c) 2016-2020, EPFL/Blue Brain Project
 """
 
 import setuptools
+import sys
+
 import versioneer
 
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -40,6 +42,9 @@
     'arbor>=0.7',
 ]
 
+if sys.version_info < (3, 8):
```
wvangeit, line 45: I wonder if we should really do a hardcore requirement?
It's very possible it would work in older versions.

anilbey: Yeah we don't have to do it.

Python 3.7 will work but it is likely that users soon will start having compatibility errors due to python3.7 drop of support. Those errors may or may not clearly state that the issue is raised because of the python version and thus confuse the users.

I guess it's taste at the end of the day. I can revert that single commit. Let me know.

AurelienJaquier (approved):

PR #448, opened 2023-03-01
Modify the way the starting centroids are generated for CMA

DrTaDa:
The goal of this PR is to avoid the over-exploration of the border of the parameter space, especially at initialisation.

### DIFF
--- a/bluepyopt/deapext/optimisationsCMA.py
+++ b/bluepyopt/deapext/optimisationsCMA.py
@@ -56,7 +56,7 @@ class DEAPOptimisationCMA(bluepyopt.optimisations.Optimisation):
         seed=1,
         offspring_size=None,
         centroids=None,
-        sigma=0.4,
+        sigma=0.3,
         map_function=None,
         hof=None,
         selector_name="single_objective",
@@ -284,7 +284,8 @@ class DEAPOptimisationCMA(bluepyopt.optimisations.Optimisation):
                 sigma=self.sigma,
                 max_ngen=max_ngen,
                 IndCreator=self.toolbox.Individual,
-                RandIndCreator=self.toolbox.RandomInd,
+                problem_size=self.problem_size,
+                ind_size=self.ind_size,
                 map_function=self.map_function,
                 use_scoop=self.use_scoop,
             )
--- a/bluepyopt/deapext/utils.py
+++ b/bluepyopt/deapext/utils.py
@@ -83,6 +83,40 @@ class WSListIndividual(list):
         super(WSListIndividual, self).__init__(*args, **kwargs)
 
 
+def generate_starters(
+    problem_size,
+    ind_size,
+    lambda_,
+    clip=True,
+    clip_threshold=0.8
+):
+    """Generate a new set of starter individual uniformly distributed in a
+    hypercube centered at zero and of length 2 * clip_threshold. Used by SO
+    and MO-CMA
+
+    Args:
+         problem_size (int): dimension of the parameter space
+         ind_size (int): number of objectives
+         lambda_ (map): population size.
+         clip (bool): should the value be clipped to the clip_threshold value
+         clip_threshold (float): value to clip the parameters to in the
+            normalized space [-1;1]
+    """
+
+    ubound = clip_threshold if clip else 1.
+    lbound = - ubound
+
+    starters = [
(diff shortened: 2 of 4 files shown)


bluepyopt/deapext/utils.py
```diff
@@ -83,6 +83,23 @@ def __init__(self, *args, **kwargs):
         super(WSListIndividual, self).__init__(*args, **kwargs)
 
 
+def generate_starters(ind_size, lambda_, clip=True, clip_threshold=0.8):
+    """Generate a new set of starter individual uniformaly distributed in a
+    hypercube centered at zero and of length 2. * clip_threshold. Used by SO
```
AurelienJaquier, line 88: Could you remove the point after the 2? For readability

bluepyopt/deapext/utils.py
```diff
@@ -83,6 +83,23 @@ def __init__(self, *args, **kwargs):
...
+    and MO-CMA"""
+
+    ubound = clip_threshold if clip else 1.
```
AurelienJaquier, line 91: Why do we initialize it between -1 and 1? Why not between parameter bounds?

bluepyopt/deapext/utils.py
```diff
@@ -83,6 +83,23 @@ def __init__(self, *args, **kwargs):
...
+    lbound = - ubound
+
+    starters = [
+        WSListIndividual(
+            numpy.random.uniform(lbound, ubound, ind_size), obj_size=ind_size)
+        for i in range(lambda_)
+    ]
+
+    return starters
+
```
AurelienJaquier, line 101: Overall what's the reason for changing the initialisation? Did the overexploring of the bounds lead to a bad exploration of the overall space? Or do we expect the bounds to not be good parameter values in general?

bluepyopt/deapext/utils.py
```diff
@@ -83,6 +83,31 @@ def __init__(self, *args, **kwargs):
...
+def generate_starters(
+    problem_size,
+    ind_size,
+    lambda_,
+    clip=True,
+    clip_threshold=0.8
+):
+    """Generate a new set of starter individual uniformaly distributed in a
+    hypercube centered at zero and of length 2. * clip_threshold. Used by SO
+    and MO-CMA"""
```
AurelienJaquier, line 95: Also could be nice to describe the arguments in the docstring here

bluepyopt/deapext/optimisationsCMA.py
```diff
@@ -211,10 +211,7 @@ def setup_deap(self):
         )
 
         # Register the evaluation function for the individuals
-        self.toolbox.register(
-            "evaluate",
-            self.evaluator.init_simulator_and_evaluate_with_lists
-        )
+        self.toolbox.register("evaluate", self.evaluator.evaluate_with_lists)
```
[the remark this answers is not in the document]
DrTaDa: Ah yes sorry, that's a mistake

bluepyopt/deapext/utils.py
DrTaDa: The parameter space used by CMA is normalized

bluepyopt/deapext/utils.py
DrTaDa: Both. The bounds are being over-explored because the parameter sets are brought back to the bounds when they are spawned outside.
And generally, our bounds are a bit "extreme" compare to the expected values and thus the solution is likely not on the bounds (but most of the models that crash are there).

DrTaDa: Note that, here if the god solution is in the missing 20%, we do not prevent the centroid to go there during the optimization.

AurelienJaquier (approved):
Great! Looks good to me

wvangeit:
So this will be a change that breaks backward compatibility, right? In essence I want to avoid that all cost in BluePyOpt. I guess now it could be acceptable because I assume the use of CMA is still limited to very few people. 
If we do it, we should at least update the version number of bluepyopt so that the changed version could be 'easily' recognized by users when needed.

DrTaDa:
Ok. I will do that.
I put the PR back as a draft because I am still investigating if these changes are really worth it.

DrTaDa:
Not relevant anymore

PR #450, opened 2023-03-14
New stagnation stopping criterion

AurelienJaquier:
For reference, here is a description of the new stopping criterion:
The problem we had with the previous stopping criterion I proposed was that, due to noise in the optimisation best fitness curve, it sometimes triggered too soon. This is why I decided to combine the same stopping criterion, **stop if the median of the last 20 values of the best fitness has improved less than 1% over the last 100 generations**, with another one, **stop if the standard deviation of the best fitness across the last 20 generations is below a threshold of 2% of the best fitness**. This prevents the first criterion to trigger due to noise, since the noise will prevent the second criterion to trigger.
Combining these two criteria will greatly decrease the number of false positive.

### DIFF
--- a/bluepyopt/deapext/CMA_SO.py
+++ b/bluepyopt/deapext/CMA_SO.py
@@ -59,6 +59,7 @@ class CMA_SO(cma.Strategy):
         RandIndCreator,
         map_function=None,
         use_scoop=False,
+        use_stagnation_criterion=True,
     ):
         """Constructor
 
@@ -74,6 +75,8 @@ class CMA_SO(cma.Strategy):
             map_function (map): function used to map (parallelize) the
                 evaluation function calls
             use_scoop (bool): use scoop map for parallel computation
+            use_stagnation_criterion (bool): whether to use the stagnation
+                stopping criterion on top of the maximum generation criterion
         """
 
         if offspring_size is None:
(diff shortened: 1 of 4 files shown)


bluepyopt/deapext/CMA_SO.py
```diff
@@ -31,7 +31,7 @@
 
 from .stoppingCriteria import (
     MaxNGen,
-    Stagnation,
```
wvangeit, line 34: What happened with Stagnation 'v1'?

AurelienJaquier: It still exists, but it was dependent on the evolution of median of the population, which was quite noisy, and not very reliant. Also it tended to trigger very late or even not at all, which made it not very good for our use cases. So I took the liberty to use the new stagnation criterion instead of the 'old' one.

AurelienJaquier:
Don't merge this yet, I was required to allow the user to not use the stagnation criterion if he/she wants to.

DrTaDa:
Looks nice !
I still feel it stops a bit early in some cases but I am not sure I would be able to do better "by hand".

DrTaDa (approved):

AurelienJaquier:
@DrTaDa are you ok with the last changes I made? Is it ok to merge this?

DrTaDa:
All good for me ! Go for it

PR #454, opened 2023-05-01
Fix get_context of multiprocessing for windows compatibility

DrTaDa:
Fixes issue #453 ?


AurelienJaquier:
Then windows users might encounter this error: #424

AurelienJaquier:
Maybe at least give a warning telling the user that things might break if s/he uses non-default dt value?

AurelienJaquier (approved):

PR #455, opened 2023-05-03
use with context manager when reading/writing files

AurelienJaquier:
To be sure that files are closed once we are finished reading / writing them.


anilbey (approved):
Thanks!

Issue #457, opened 2023-05-09
BluePyOpte error during TM-fitting


NataliBarros:
I would like to replicate the multiple frequency TM-fitting for my data as it is done in here:
https://github.com/BlueBrain/BluePyOpt/blob/master/examples/tsodyksmarkramstp/tsodyksmarkramstp_multiplefreqs.ipynb

However when I run the code in my venv I get the following error when running the second cell with code:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [3], in <cell line: 13>()
     11 # Create BluePyOpt optimization and run 
     12 evaluator = tmevaluator.TsodyksMarkramEvaluator(data, optconf)
---> 13 opt = bpop.optimisations.DEAPOptimisation(evaluator, offspring_size=500, map_function=pool.map,
     14                                           eta=20, mutpb=0.3, cxpb=0.7)
     15 pop, hof, log, history = opt.run(max_ngen=50)
     17 # Get best individual

File ~/venv_jupyterhub/lib/python3.8/site-packages/bluepyopt/deapext/optimisations.py:146, in DEAPOptimisation.__init__(self, evaluator, use_scoop, seed, offspring_size, eta, mutpb, cxpb, map_function, hof, selector_name)
    143 # Create a DEAP toolbox
    144 self.toolbox = deap.base.Toolbox()
--> 146 self.setup_deap()

File ~/venv_jupyterhub/lib/python3.8/site-packages/bluepyopt/deapext/optimisations.py:204, in DEAPOptimisation.setup_deap(self)
    194 self.toolbox.register(
    195     "population",
    196     deap.tools.initRepeat,
    197     list,
    198     self.toolbox.Individual)
    200 # Register the evaluation function for the individuals
    201 # import deap_efel_eval1
    202 self.toolbox.register(
    203     "evaluate",
--> 204     self.evaluator.init_simulator_and_evaluate_with_lists
    205 )
    207 # Register the mate operator
    208 self.toolbox.register(
    209     "mate",
    210     deap.tools.cxSimulatedBinaryBounded,
    211     eta=ETA,
    212     low=LOWER,
    213     up=UPPER)

AttributeError: 'TsodyksMarkramEvaluator' object has no attribute 'init_simulator_and_evaluate_with_lists'
```
Can anybody help me with this issue? Thanks.

wvangeit:
[...]

PR #458, opened 2023-05-09
Fix tsodyksmarkramstp example

AurelienJaquier:
Solves Issue #457


DrTaDa (approved):
So the method is called init_simulator_and_evaluate_with_lists but it does not actually init a simulator ? haha

AurelienJaquier:
Yeah. it is called during optimisation, assuming the evaluator is a Cellevaluator, which has a simulator and this method. So for the other evaluators, we have to create this function, even if it is just calling evaluate_with_lists.

PR #460, opened 2023-05-10
possibility of using a minimum amount of time before saving optimisation

AurelienJaquier:
We already have the cp_frequency that allows us to skip a given number of generations when saving the checkpoint.
For the case where we don't know beforehand how much time does a generation take, and we don't want to register all the generations if they are computed fast, we can use this new option.

Is not used by default

### DIFF
--- a/bluepyopt/deapext/algorithms.py
+++ b/bluepyopt/deapext/algorithms.py
@@ -26,6 +26,7 @@ import random
 import logging
 import shutil
 import os
+import time
 
 import deap.algorithms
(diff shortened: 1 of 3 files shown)


DrTaDa:
The goal is to reduce the I/O when the evaluation is fast ?

AurelienJaquier:
> The goal is to reduce the I/O when the evaluation is fast ?

Yes

DrTaDa (approved):

PR #462, opened 2023-06-06
remove ineffective checks against python2

anilbey:


anilbey:
@wvangeit I think only you have the authority to merge

wvangeit:
Is there any specific reason to remove these lines, apart from readability?
I don't really feel like breaking compatibility just for the sake of breaking compatibility. I'm just worried there might be some users using notebooks on old systems etc.

anilbey:
No, only readability and a bit of confusion :)
E.g. `from __future__` is not the future but the past now

wvangeit:
Ok, but I don't consider that enough arguments to possibly break external user's scripts.

anilbey:
Ok, sure. I'm closing.

PR #465, opened 2023-07-11
tag and release at each push to master

AurelienJaquier:


AurelienJaquier:
@wvangeit should I also add a .zenodo.json file to BluePyOpt?

wvangeit (approved):

PR #473, opened 2023-10-05
remove __future__ imports

AurelienJaquier:
I only removed one line in each jupyter notebook, and it does not look like anything else has changed, when checking visually. @anilbey do you know why other lines have been unilateraly changed here?

### DIFF
--- a/bluepyopt/deapext/tools/selIBEA.py
+++ b/bluepyopt/deapext/tools/selIBEA.py
@@ -1,7 +1,5 @@
 """IBEA selector"""
 
-from past.builtins import xrange  # pylint: disable=W0622
-
 """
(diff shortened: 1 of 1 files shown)


anilbey:
Nice change.

Yeah jupyter and git do these. Those differences must be created after you execute the cells.

anilbey:
So I think if you git reset and then remove the lines and not execute the notebook cells then it should not add those.

AurelienJaquier:
I actually did not execute the cells. I just removed the line, and saved in VScode.

anilbey:
Oh, I see. Hmm if vscode did something and not displayed it, then the javascript lines can still be selected via `git add -p PATH_TO_NOTEBOOK` if you want to remove them.

PR #486, opened 2024-02-02
update jinja template to have getCell function

AurelienJaquier:
needed for hoc to be loadable in BlueCelluLab

also tested BluePyMM tests with this BluePyOpt branch, and everything passed


anilbey (approved):
Looks great, thanks

PR #493, opened 2024-04-25
adding build block to .readthedocs.yml

AurelienJaquier:
readthedocs is failing because it expects a build block.


AurelienJaquier:
`macos-latest` has recently been updated to version `14`, and this creates problems with our workflow, so I reverted to what used to be `macos-latest` before the update: `macos-12`.

ilkilic (approved):

PR #498, opened 2024-08-20
fix arbor tests with latest arbor version

AurelienJaquier:
dropped support for python 3.8 because latest arbor does not support it and python 3.8 will reach end of life in one month.


ilkilic (approved):

bluepyopt/ephys/templates/acc/decor_acc_template.jinja2
```diff
@@ -10,7 +10,7 @@
     {%- endif %}
     {%- else %}
     {%- for param in params %}
-    (default ({{ param.name }} {{ param.value }}))
+    (default ({{ param.name }} {{ param.value }} (scalar 1.0)))
```
thorstenhater, line 13: I won't check all of those in detail, but this is correct for a fixed value.
If the tests are green, I'd assume all are correct.

bluepyopt/ephys/templates/acc/decor_acc_template.jinja2
thorstenhater, line 1: As said before using Arbor to write these might be an easier option.

thorstenhater (commented):
I'll leave these as comments: most is as expected, but I have some comments here and there.

bluepyopt/ephys/morphologies.py
```diff
@@ -307,6 +307,8 @@ def load(morpho_filename, replace_axon):
             morpho = arbor.load_component(morpho_filename).component
         elif morpho_suffix == '.swc':
             morpho = arbor.load_swc_arbor(morpho_filename)
+            # turn loaded_morphology into morphology type
+            morpho = arbor.morphology(morpho.segment_tree)
```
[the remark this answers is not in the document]
AurelienJaquier: Done in latest commit

bluepyopt/ephys/protocols.py
```diff
@@ -631,7 +636,7 @@ def instantiate_recordings(self, cell_model, use_labels=False):
         """Instantiate recordings"""
 
         # Attach voltage probe sampling at 10 kHz (every 0.1 ms)
-        for i, rec in enumerate(self.recordings):
+        for _, rec in enumerate(self.recordings):
```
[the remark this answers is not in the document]
AurelienJaquier: Keeping the index in the latest commit

bluepyopt/ephys/protocols.py
```diff
@@ -648,7 +653,9 @@ def instantiate_recordings(self, cell_model, use_labels=False):
 
             cell_model.probe('voltage',
                              arb_loc.ref if use_labels else arb_loc.loc,
-                             frequency=10)  # could be a parameter
+                             "0",  # tag: default is '0'
```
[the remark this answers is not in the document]
AurelienJaquier: Done in latest commit.

bluepyopt/ephys/templates/acc/decor_acc_template.jinja2
AurelienJaquier: The tests are green

bluepyopt/ephys/templates/acc/decor_acc_template.jinja2
AurelienJaquier: Is there a function that can write these kind of files?

thorstenhater: Yes, there's versions for full cable cells or just parts thereof.
https://docs.arbor-sim.org/en/latest/fileformat/cable_cell.html#parsable-arbor-components-and-meta-data
In Python, you'd construct a cable cell (or the parts you're interested in) and call `arbor.write_component(thing, filename)`
https://docs.arbor-sim.org/en/latest/python/cable_cell_format.html#reading-and-writing-arbor-components

thorstenhater:
Picking one of the examples from the test suite
```LISP
(arbor-component
  (meta-data (version "0.9-dev"))
  (decor
    (paint (region "soma") (membrane-capacitance 0.01 (scalar 1.0)))
    (paint (region "soma") (density (mechanism "default::hh" ("gnabar" 0.10299326453483033) ("gkbar" 0.027124836082684685))))))
```
should be produced by
```py
import arbor as A

d = A.decor()
d.paint('"soma"', cm=0.01)
d.paint('"soma"', A.density('default:hh', gnabar=0.10299326453483033, gkbar=0.027124836082684685))

A.write_component(d, 'test.acc')
```

bluepyopt/ephys/templates/acc/decor_acc_template.jinja2
lukasgd: I'd recommend to additionally check the major examples' notebooks qualitatively by inspection (not everything is run as part of the test suite).

bluepyopt/ephys/templates/acc/decor_acc_template.jinja2
lukasgd: That would basically mean to turn this template into code at the end of the ACC exporter by instantiating Arbor objects that are then serialized again. Could be an option - I guess it comes down to how much control you need over the export vs. learning the s-expression syntax. The original design of BluePyOpt was to make model exporting configurable (overridable by the user) through templates, hence I chose this option. 

In my view, most of the complexity in the ACC exporter comes from converting BluePyOpt's frontend representation to one suitable for Arbor. That last step of dumping the decor is less complicated, but requires some understanding of the format.

I don't want to take any particular side, though. Would only suggest that if you plan to make major changes to the ACC exporter, maybe a good idea would be to do that in a separate PR.

thorstenhater: Yes, I'd support splitting this. One straight port that can be verified and a lateral move to another scheme.

> The original design of BluePyOpt was to make model exporting configurable (overridable by the user) through templates, hence I chose this option.

This is why I as an outsider should have an opinion here :D

AurelienJaquier:
Thanks for your help @lukasgd @thorstenhater , I think this PR is good to be merged.
For the ACC exporter PR, we will have to tackle it at a later date, because I have a bunch of other priorities right now and we are a little bit short on manpower.

PR #500, opened 2024-10-08
use pyproject.toml and remove versioneer

AurelienJaquier:


pyproject.toml
```diff
@@ -0,0 +1,78 @@
+[build-system]
+requires = ["setuptools >= 64", "setuptools-scm>=8.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "bluepyopt"
+authors = [
+    {name = "Blue Brain Project, EPFL", email = "werner.vangeit@epfl.ch"},
```
ilkilic, line 8: I am not sure that they can reach him through this email address anymore

pyproject.toml
```diff
@@ -0,0 +1,78 @@
...
+]
+description="Bluebrain Python Optimisation Library (bluepyopt)"
```
ilkilic, line 10: Perhaps a space between blue and brain?

pyproject.toml
AurelienJaquier: I think it is still valid because he is an external collaborator

pyproject.toml
AurelienJaquier: Good catch. Fixed in latest commit

ilkilic (approved):

